PR #48, opened 2018-05-28
PR: Harmonize body & sidebar font to 16px and Nav head to L2 head

CAM-Gerlach:
### Pull Request Checklist

* [x] Read and followed this repo's [Contributing Guidelines](https://github.com/spyder-ide/spyder/blob/master/CONTRIBUTING.md)
* [x] Based your PR on the latest version of the correct branch (master or 3.x)
* [x] Checked your writing carefully for correct English spelling, grammar, etc
* [x] Described your changes and the motivation for them below
* [x] Noted what issue(s) this pull request resolves, creating one if needed

## Description of Changes

Harmonizes the body font and main sidebar font to an even 16px, matching RTD, and bumped the sidebar subheads down to 14.5 px. Internally, this will make a the docs a tad more readable, while saving a little more space in the sidebar. Externally, it will enable easy matching with the blog. Also, bumps the "Navigation" header up just a touch to match the L2 subheads properly.

### Issue(s) Resolved

Fixes #26

PR #58, opened 2018-07-03
PR: Improve style

ccordoba12:
- Increase text size to 17px
- Increase total width to 1100px
- Increase sidebar width to 250px
- Use Raleway for headers
- Increase line-height

Note: The current style is really bad because the text looks too small and packed.


ccordoba12:
Merging to iterate quickly.

CAM-Gerlach:
The irony that this was merged instantly and unilaterally to iterate quickly when #48 , which already concerned the first three items (most notably increasing font size) and been patiently awaiting review for over a month, is not lost on me, heh...but bygones are bygones.

> Increase line height

Discussed elsewhere

> Increase text size to 17px

Discussed in other thread.

> Use Raleway for headers

Agreed. However, their font size needs to be reduced 10-20% accordingly.

The headers are currently at least 10-15% too large, due to the aforementioned fact that at a given px size, a sans serif font (like Raleway) visually looks around 10-20% larger than an equivalent serif font (like Amiri) for example, here's 12 pt Liberation Serif and 11 pt Liberation Sans;

image

And even more starkly, their equivalents in MS Word, 12 pt Times (Serif) and 11 pt Arial (Sans)

image

This, in turn, makes the body text look much smaller (and 16 px more "too small) by comparison; setting the headers to a more reasonable size (~15% smaller) will proportionally decrease how small the font looks.

> Increase total width to 1100 px

While I visually prefer the wider page with as well, the major problem with it (and the reason I didn't change it from the 960 px in the template) is that because the current docs layout isn't adaptive to screen width, the text gets hard cut off if the window is narrower than 1100 px, which is likely quite common for many users with only 1 or 2 >= 1200p (1080p, 720p, etc) monitors in horizontal orientation (since they are likely to want to split to window horizontally in order to conserve enough screen space to actually get work done, thus they only have 960 px width, as do >= 1600 x 900 and >= 1280 x 960 monitors among others when in vertical/portrait mode, as most of my programmer/data science colleagues prefer (thus the reason that the original theme sets it to be 960 px wide initially). Even a 1920 x 1080 in portrait mode will potentially cut off at least a little content.

Therefore, 1100 px width (instead of 960 px) breaks most use cases that doesn't involve the docs on a hi-dpi monitor or 2-3+ horizontal screens (where the user has the space aside from Spyder itself and any variable explorers, terminals, text editors, other webpages/docs, etc to make the Spyder docs fullscreen horizontal), since they will be cut off and thus unreadable without constant horizontal scrolling. In turn, this leaves less room for the text, thus making a smaller size more advisable to maintain a reasonable information density and overall look the same size.

> Increase sidebar width to 250px

I guess its good to have a little whitespace between the sidebar and the docs, but too much is just wasted space when it could instead be used to reduce the total width (or equivalently widen and include more doc content) to a level below which breaks usability; that ~50 px that could be recovered (particularly when coupled with a 16 px vs. 16.8/17 px font size) goes a long way toward minimizing the effect of a 140 px narrower total width to not break usability as discussed above.

ccordoba12:
> The irony that this was merged instantly and unilaterally to iterate quickly when #48 , which already concerned the first three items (most notably increasing font size) and been patiently awaiting review for over a month

Sorry, didn't see it.

> Agreed. However, their font size needs to be reduced 10-20% accordingly.

Done in PR #61.

> While I visually prefer the wider page with as well, the major problem with it (and the reason I didn't change it from the 960 px in the template) is that because the current docs layout isn't adaptive to screen width, the text gets hard cut off if the window is narrower than 1100 px

Ok, I'll decrease the total width to 1000px, and not 960px, to allow for a larger sidebar.

> I guess its good to have a little whitespace between the sidebar and the docs, but too much is just wasted space when it could instead be used to reduce the total width

Look at the output of Git books:

http://www.theia-ide.org/doc/authoring_extensions

So it's not bad to have a well separated sidebar from the main content.

CAM-Gerlach:
> Done in PR #61.

Looks great, thanks :+1: 

> Ok, I'll decrease the total width to 1000px, and not 960px, to allow for a larger sidebar.

1000 px doesn't fix the problem since since it still cuts off the edge on most of the same setups as mentioned above, and not only that but even scrolling past the edge doesn't even render the missing text. The real problem with all of this is the layout is fixed width instead of adaptive, then it would work well on all screen sizes.

image

> Look at the output of Git books: So it's not bad to have a well separated sidebar from the main content.

Yes, indeed. But the key thing that site does differently is the sidebar is pinned to the left and the content width is adaptive; i.e. if the window is narrower than the nominal width, it will soft-wrap rather than just cutting stuff off. This would solve both of these problems much more cleanly and elegantly than making the main content narrower, and we could even make it 1200 px or something nominal and add much more space between the sidebar and the content (or even pin it to the left). Is that easy to do in CSS?

ccordoba12:
> Yes, indeed, a layout like that site would be ideal.

Then please open an issue about it to fix in Javascript.

CAM-Gerlach:
Opened as #63

PR #59, opened 2018-07-03
PR: Remove all images in Overview

ccordoba12:
Those images were really, really confusing because they display something that doesn't look at all like Spyder.

Note: This doesn't admit discussion @CAM-Gerlach, sorry.


CAM-Gerlach:
...

PR #61, opened 2018-07-03
PR: More style changes

ccordoba12:
- Decrease total width to 1000px
- Decrease line-height to 1.57
- Decrease main font-size to 16.8px
- Decrease header font-size by 20%
- Fix style of links on hover

### DIFF
--- a/doc/_static/custom.css
+++ b/doc/_static/custom.css
@@ -39,5 +39,5 @@ ul.simple {
 }
 
 div.body p, div.body dd, div.body li {
-    line-height: 1.9em;
+    line-height: 1.7em;
 }
--- a/doc/_static/custom.css
+++ b/doc/_static/custom.css
@@ -4,12 +4,12 @@ cite::before {
 
 cite {
     display: inline;
-    font-size: 17px;
+    font-size: 16.8px;
     color: black;
 }
 
 #doc-content{
-    font-size: 17px;
+    font-size: 16.8px;
 }
 
 .logo {
--- a/doc/_static/custom.css
+++ b/doc/_static/custom.css
@@ -4,12 +4,12 @@ cite::before {
 
 cite {
     display: inline;
-    font-size: 18px;
+    font-size: 17px;
     color: black;
 }
 
 #doc-content{
-    font-size: 18px;
+    font-size: 17px;
 }
 
 .logo {
--- a/doc/_static/custom.css
+++ b/doc/_static/custom.css
@@ -4,12 +4,12 @@ cite::before {
 
 cite {
     display: inline;
-    font-size: 17px;
+    font-size: 18px;
     color: black;
 }
 
 #doc-content{
-    font-size: 17px;
+    font-size: 18px;
 }
 
 .logo {
@@ -39,5 +39,5 @@ ul.simple {
 }
 
 div.body p, div.body dd, div.body li {
-    line-height: 1.7em;
+    line-height: 1.57em;
 }
--- a/doc/_static/custom.css
+++ b/doc/_static/custom.css
@@ -41,3 +41,7 @@ ul.simple {
 div.body p, div.body dd, div.body li {
     line-height: 1.57em;
 }
+
+a:hover {
+    text-decoration: none;
+}
--- a/doc/_static/custom.css
+++ b/doc/_static/custom.css
@@ -45,3 +45,21 @@ div.body p, div.body dd, div.body li {
 a:hover {
     text-decoration: none;
 }
+
+div.body h1 {
+    margin-top: 0;
+    padding-top: 0;
+    font-size: 220%;
+}
+
+div.body h2 {
+    font-size: 160%;
+}
+
+div.body h3 {
+    font-size: 130%;
+}
+
+div.body h4 {
+    font-size: 110%;
+}
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -138,7 +138,7 @@ html_theme_options = {
         },
     'sidebar_collapse': True,
     'show_related': True,
-    'page_width': '1100px',
+    'page_width': '1000px',
     'sidebar_width': '250px',
     'fixed_sidebar': False,
     'gray_1': '#303030',
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -146,7 +146,7 @@ html_theme_options = {
     'gray_3': '#EE1C24',
     'font_family': "Raleway, 'Open Sans', Arial, sans-serif",
     'head_font_family': "Raleway, 'Open Sans', Arial, sans-serif",
-    'font_size': '17px',
+    'font_size': '18px',
     'link': '#EE1C24',
     'link_hover': '#EE1C24'
     }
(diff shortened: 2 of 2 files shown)


ccordoba12:
@CAM-Gerlach, do you agree with `17px` or `18px` for the main font size?

CAM-Gerlach:
Personally, I prefer 16 px (as I'd set it to in #48 ) and 1.57 (or less, but I'm not gonna argue :) since 18 px is IMO too large for docs, which have a general need to communicate information more densely than a blog post, but we want it to be at least a little more distinct from the blog if we're not going to use exactly the same and it matches the footer text (a general design axiom my mom frequently reminds me of is to not have fonts only 1-2 px apart in size in a layout). Furthermore, my general impression from our peer doc sites is that the text size (and line spacing, but I think 1.57 is okay for now) is generally smaller than that for their respective blogs, but I can do a quick survey of the same 10 sites as before to give you an idea if that's the case.

Alternatively, we could harmonize both the blog and docs at 17 px and be done with it, but I tend think you're right that they are better off distinct, as long as they aren't *too* close to one another.

ccordoba12:
> I prefer 16 px

Could we agree on 17px? I think 16px is too small.

> 18 px is IMO too large for docs

Agreed.

CAM-Gerlach:
> Could we agree on 17px? I think 16px is too small.

16.8 px would match our current sidebar font size and doesn't look *that* big as it is, so I would conditionally agree as things stand now. However, aside from the fact that it would require the "Connect with us" in the footer to be increased to that size, several potential changes will make 17 px look bigger (and 16 px less small) as I was planning to explain on #58 : 

The headers are currently at least 10-15% too large, due to the aforementioned fact that at a given px size, a sans serif font (like Raleway) visually looks around 10-20% larger than an equivalent serif font (like Amiri) for example, here's 12 pt Liberation Serif and 11 pt Liberation Serif;

image

And even more starkly, their equivalents in MS Word, 12 pt Times and 11 pt Arial 

image

This, in turn, makes the body text look much smaller (and 16 px more "too small) by comparison; setting the headers to a more reasonable size (~15% smaller) will proportionally decrease how small the font looks.

Also, while I visually prefer the wider page with as well, the major problem with it (and the reason I went with it initially) is that because the current docs layout isn't adaptive to screen width, the text gets hard cut off if the window is narrower than 1100 px, which is likely quite common for many users with only 1 or 2 >= 1200p (1080p, 720p, etc) monitors in horizontal orientation (since they are likely to want to split to window horizontally in order to conserve enough screen space to actually get work done, thus they only have 960 px width, as do >= 1600 x 900 and >= 1280 x 960 monitors among others when in vertical/portrait mode, as most of my programmer/data science colleagues prefer (thus the reason that the original theme sets it to be 960 px wide initially). Even a 1920 x 1080 in portrait mode will potentially cut off at least a little content.

Therefore, 1100 px width (instead of 960 px) breaks most use cases that doesn't involve the docs on a hi-dpi monitor or 2-3+ horizontal screens (where the user has the space aside from Spyder itself and any variable explorers, terminals, text editors, other webpages/docs, etc to make the Spyder docs fullscreen horizontal), since they will be cut off and thus unreadable without constant horizontal scrolling. In turn, this leaves less room for the text, thus making a smaller size more advisable to maintain a reasonable information density and overall look the same size.

CAM-Gerlach:
Okay, here you go (body/h2 headers—we're currently at 30.6 px, with line spacing in parenthesis) :

Even without adjusting for serifs, the mean size is 15.2 px, with a median of 15.5 px, and none with font size larger than 16 px (except for one with 17 px, but that's a serif font thus equivalent to a 14.5-16 px sans). (Also, incidentally, the mean heading size without adjustment was 26.8px, median 26.1 px, with only 3 of 10 sans faces above 27 px; the mean and median line spacing was 1.46 px with only 2 of 1.57 or above). 

Therefore, at least judging from all checked peers the case for 16 px seems very compelling. Additionally, heading size could be bumped down at least a little, although since its presumably based on ``1.8 em`` it *should* automatically follow the decreased base font size down to at least 28.8 px, which is at least somewhat more reasonable (though 1.7 em, at 27.2 px would be even better, and still above average; meanwhile, we could also go with 1.5 line spacing to be more consistent with others, if we're not going to match the website in size or font, etc., though I won't make a fuss over it if we don't). 

Full results:

* 16 px sans / 24 px sans [Readthedocs theme](https://docs.readthedocs.io/en/latest/) (1.5)
* 17 px serif / 30.6 px serif [Alabaster doc theme](https://alabaster.readthedocs.io/en/latest/) (1.4)
* 14 px sans / 21 px sans [Sphinx docs](http://www.sphinx-doc.org/en/master/usage/installation.html) (~1.2)
* 15 px sans / 32 px sans [Jupyter/IPython docs](https://jupyter.readthedocs.io/en/latest/ipython/content-ipython.html) (1.4)
* 14 px sans / 23 px sans [Rstudio docs](https://support.rstudio.com/hc/en-us/articles/115002344588-Configuration-and-sizing-recommendations) (1.64) 
* 14 px sans / 27 px sans [Yhat Rodeo docs](http://rodeo.yhat.com/docs/) (1.5)
* 16 px sans / 25.6 px sans [Enthought Canopy IDE docs](https://docs.enthought.com/canopy/2.1/quick-start/install_macos.html) (1.3)
* 16 px sans / 36 px sans [Pycharm docs](https://www.jetbrains.com/help/pycharm/meet-pycharm.html)(1.56)
* 16 px sans / 19.2 px sans [ATOM docs](https://flight-manual.atom.io/getting-started/sections/why-atom/) (1.68)
* 14 px sans / 30 px sans [OpenStack Docs](https://docs.openstack.org/swift/queens/install/controller-install-obs.html) (1.42)

ccordoba12:
> 16.8 px would match our current sidebar font size and doesn't look that big as it is, so I would conditionally agree as things stand now

Ok, 16.8px will be.

ccordoba12:
Merging because this improves the current look-and-feel a lot. The only controversial thing is total width, but that be discussed further.

CAM-Gerlach:
> Merging because this improves the current look-and-feel a lot.


> The only controversial thing is total width, but that be discussed further.

Did you see my followup post, where I found that not a single peer site used a font size above 16 px for their docs, which along with the other benefits of the smaller size due to the changes makes a rather compelling case for 16 px?

Other than the font size, yes, and I guess that's indirectly in connection with the total width so we can consider it with that perhaps. 1000 px doesn't fix the problem since since it still cuts off the edge on most setups, and not only that but even scrolling past the edge still even render the missing text. The real problem with all of this is the layout is fixed width instead of adaptive, then it would work well on all screen sizes.

image

ccordoba12:
> Did you see my followup post, where I found that not a single peer site used a font size above 16 px for their docs

Yes, and I mentioned Git books as a counterexample.

ccordoba12:
Sorry, I thought I did (I mentioned in connection to the sidebar). But it's a real counterexample.

CAM-Gerlach:
> Yes, and I mentioned Git books as a counterexample.

The only place I see that mentioned is on a different issue thread and in response to separate issue (the sidebar width). Furthermore, the site you linked indeed has a 16 px font size, exactly as I'm advocating for, so I'm not sure how that's in any way a counterexample.

ccordoba12:
It's probably 16px but it doesn't look like it. Sorry, I stand by 17px for now.

ccordoba12:
And by the way, they use 1.7 for their line-height :-)

CAM-Gerlach:
> It's probably 16px but it doesn't look like it. Sorry, I stand by 17px for now.

Alternative facts :)

PR #45, opened 2018-05-23
PR: Add documentation and images for Profiler module

CAM-Gerlach:
### Pull Request Checklist

* [x] Read and followed this repo's [Contributing Guidelines](https://github.com/spyder-ide/spyder/blob/master/CONTRIBUTING.md)
* [x] Based your PR on the latest version of the correct branch (master or 3.x)
* [x] Checked your writing carefully for correct English spelling, grammar, etc
* [x] Described your changes and the motivation for them below
* [x] Noted what issue(s) this pull request resolves, creating one if needed

## Description of Changes

Adds a decent baseline documentation page for the Profiler module and relevant images.

Aside from adding the page itself, it also needs to be added a few other places, which will be done once #44 gets merged; this PR should accordingly not be merged before that.

* [x] Add new Profiler doc
    * [x] Add text content
    * [x] Add images
    * [x] Format to style guide
* [ ] Add links elsewhere
    * [ ] To ``:toctree:`` in ``index``
    * [ ] Under Related Components in ``Static Code Analysis``
    * [ ] Inline under Profiler bullet in ``Overview``
    * [ ] Inline in other docs if needed?

### Issue(s) Resolved

Fixes #41

PR #29, opened 2018-05-07
PR: Eliminate redundant links from sidebar, add standard links to readme and fix minor layout bugs

CAM-Gerlach:
### Pull Request Checklist

* [x] Read and followed this repo's [Contributing Guidelines](https://github.com/spyder-ide/spyder/blob/master/CONTRIBUTING.md)
* [x] Based your PR on the latest version of the correct branch (master or 3.x)
* [x] Checked your writing carefully for correct English spelling, grammar, etc
* [x] Described your changes and the motivation for them below
* [x] ~~Noted what issue(s) this pull request resolves, creating one if needed~~

## Description of Changes

Currently, our Facebook and especially Twitter links aren't really linked from anywhere but our OpenCollective to my knowledge, and correspondingly have very low SEO rankings and are very hard to find. Therefore, this PR adds them to our readme and sidebar, and fixes a very minor issue with the ToC link layout (along the lines of #27 ).

### Issue(s) Resolved

Minor meta-task; not needed.

PR #62, opened 2018-07-03
PR: Remove unnecessary sidebar links

ccordoba12:
The sidebar is currently too crammed. This is my solution to remove unnecessary stuff.

Links removed:

- Main Website: It can be accessed from the top navbar. No need to repeat it.
- Download (Anaconda): Users can go to the Installation section
- Spyder Github: No documentation can be gained by going to it.
- Donate: The Donate button needs to make part of the footer (@dalthviz, please make sure of that).
- Relations: Our docs are pretty flat right now, so I don't see the need of it (it could be reactivated in Spyder 4)-

Links changed:

- Dev Wiki to Wiki: I think it's better to point out our users to our main wiki.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -191,7 +191,7 @@ html_sidebars = {
     '**': [
         'about.html',
         'navigation.html',
-        'relations.html',
+        #'relations.html',
         'searchbox.html',
         'donate.html',
     ]
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -126,15 +126,11 @@ html_theme_options = {
     'travis_button': False,
     'codecov_button': False,
     'extra_nav_links': {
-        'Main Website': 'https://www.spyder-ide.org/',
-        'Download (Anaconda)': 'https://www.anaconda.com/download/',
-        'Spyder Github': 'https://github.com/spyder-ide/spyder',
         'Troubleshooting': ('https://github.com/spyder-ide/spyder/wiki/'
                             + 'Troubleshooting-Guide-and-FAQ'),
         'Dev Wiki': 'https://github.com/spyder-ide/spyder/wiki/Dev:-Index',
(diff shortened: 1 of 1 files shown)


CAM-Gerlach:
I already did all of this (except for maybe removing the relations) in #29 as we discussed. That was the original basis for the footer, if you recall.

ccordoba12:
That PR has other things going on (like several style changes).

ccordoba12:
Besides, we disagree on what should be removed.

CAM-Gerlach:
> Besides, we disagree on what should be removed.

Perhaps we disagree on that we disagree :) Main website's a no brainer, and with the new footer, the Github, Gitter Chatroom and Google Group can all be removed, I have no objection to relations being removed, and its fine to direct them to our wiki I guess. The download and donate links can be removed here but I strongly feel they should go in the common top bar, rather than bury them in the footer, to make them way more visible and prominent since they are the primary two things we want to direct users to do to help themselves and the project; otherwise we're shooting ourselves in the foot on user adoption and funding and well as frustrating our users who generally expect them there, as they are on most similar sites.

CAM-Gerlach:
This PR currently still has Gitter Chatroom and Google Groups in the sidebar, which are both in the footer, embedded as a button on the page (Gitter), and less relevant than Github and the others. Presumably, they should be removed as well.

CAM-Gerlach:
See spyder-ide/spyder-docs#45 where I incorporated the remaining changes here such as that PR is now a superset of this one (eliminating all the things you do, plus Gitter and Google Groups since those are already in the footer and better belong there), as well as a few minor and hopefully uncontroversial semi-related fixes. If you agree, this PR can be closed and that one reviewed and accepted.

ccordoba12:
> See spyder-ide/spyder-docs#45 where I incorporated the remaining changes here such as that PR is now a superset of this one

Great! Closing this one then.

Issue #63, opened 2018-07-03
Make content width adaptive to window size instead of cutting off text


CAM-Gerlach:
## Issue Description

Currently, our docs content (and sidebar) is a fixed width, which results in hard cutting off of the text in many scenarios, and even worse, scrolling past the edge doesn't render the missing text. This forces us to either exclude large classes of user setups/use cases from being able to view the docs properly, or use an unreasonably narrow content width.

![image](https://user-images.githubusercontent.com/17051931/42246426-7fe58d6e-7eea-11e8-93da-b3fd9c753ffd.png)

The solution is to make the width of the docs content (and perhaps the sidebar, to a lesser extent) adaptive; i.e. starting at 1200 px max total width and then reducing the width and soft-wrapping the text if the window is smaller than that; also, we can pin the sidebar to one side to make the margins more even. Basically, something like this [Github books output](http://www.theia-ide.org/doc/authoring_extensions) mentioned by @ccordoba12 .

Hopefully its possible to do in CSS but if not, then some simple JS perhaps. I'm guessing @dalthviz would be the one to do this?

ccordoba12:
@dalthviz, please work on this one. Take a look at how Github books work, as @CAM-Gerlach suggested.

Also, this probably requires you to add some additional Javascript to the output generated by Sphinx. I mean, I don't think this could be done with CSS alone (but I could be wrong).

CAM-Gerlach:
This is currently more serious than initially though, since right now thanks to the increases in the sidebar and total content width, the docs are cut off and essentially unreadable if the window width is less than 1100 px (on a portrait-format display common in software development, when split to one half of a 1080p monitor, on older monitors, on mobile devices, etc,) with **no** scrollbar and thus no ability to even scroll to see the missing portions on either Chrome or Firefox; the above is just how it looks when rendered locally but on our actual site there is no scrollbar at all. 

This is a major issue from a usability perspective, completely breaking many likely desired use-cases. A temporary fix to at least make it usable in most of these scenarios would be reducing the total width back to 900 or 960 px as I had it (for exactly this reason), but something like what @ccordoba12 suggested with Git Books is a far better permanent solution all around.

CAM-Gerlach:
Fixed by PR #127

PR #68, opened 2018-07-08
PR: Tweak sidebar font size and line height to match revised doc content style

CAM-Gerlach:
### Pull Request Checklist

* [x] ~~Checked your writing carefully for correct English spelling, grammar, etc~~
* [x] ~~Noted what issue(s) this pull request resolves, creating one if needed~~

## Description of Changes

The content body, header, and line spacing font parameters were changed in the recent series of PRs superseding #48 ; however, unlike in that PR, the corresponding sidebar sizes were not adjusted accordingly. Therefore, this PR ports over those changes—namely, setting the proper body, heading, and line-height sizes in the sidebar relative to the body text—as well as taking the opportunity to do a bit of minor cosmetic cleanup and consolidation in ``custom.css``.

### Issue(s) Resolved

N/A; minor followup to earlier PRs

### DIFF
--- a/doc/_static/custom.css
+++ b/doc/_static/custom.css
@@ -8,8 +8,9 @@ cite {
     color: black;
 }
 
-#doc-content{
+#doc-content {
     font-size: 16px;
+    padding-bottom: 2em;
 }
 
 .logo {
@@ -75,11 +76,7 @@ div.body h4 {
     font-size: 110%;
 }
 
-#doc-content{
-    padding-bottom: 2em;
-}
-
-.pipe-red{
+.pipe-red {
     color: red;
     font-weight: bold;
 }
@@ -88,7 +85,7 @@ div.body h4 {
     font-size: 15px;
 }
 
-#fh5co-footer .container .row .text-center p small{
+#fh5co-footer .container .row .text-center p small {
     font-size: 16px;
     line-height: 2.8em;
 }
--- a/doc/_static/custom.css
+++ b/doc/_static/custom.css
@@ -18,10 +18,21 @@ cite {
 
 div .sphinxsidebarwrapper {
     padding-top: 10px;
+    font-size: 16px;
+    line-height: 1.57em;
 }
 
 div .sphinxsidebarwrapper h3 {
     margin: 0px;
+    padding: 0px;
+}
+
+div .sphinxsidebar ul li.toctree-l1 > a {
+    font-size: 100%;
+}
+
+div .sphinxsidebar ul li.toctree-l2 > a {
+    font-size: 95%;
 }
 
 kbd {
@@ -52,7 +63,7 @@ div.body h1 {
     font-size: 220%;
 }
 
-div.body h2 {
+div.body h2, div .sphinxsidebarwrapper h3 {
     font-size: 160%;
 }
 
--- a/doc/_static/custom.css
+++ b/doc/_static/custom.css
@@ -68,7 +68,7 @@ div.body h2, div .sphinxsidebarwrapper h3 {
     font-size: 160%;
 }
 
-div.body h3 {
+div.body h3, div.admonition p.admonition-title {
     font-size: 130%;
 }
 


ccordoba12:
>  however, unlike in that PR, the corresponding sidebar sizes were not adjusted accordingly

It seems you're trying to make the sidebar to have the same font size and other features of the main text. If that's the case, I don't agree with it because it doens't look good.

I think it's fine for the sidebar to have a smaller font than the main text so users can clearly see the difference between the two.

CAM-Gerlach:
> If that's the case, I don't agree with it because it doens't look good.

What about it doesn't look good, specifically? The l2 links are too small? Navigation heading's too large? Line spacing's too narrow?

Obviously its somewhat a matter of one's own taste in design, but comparing between looks of the current mainline and this branch and ignoring the objective considerations below, my eye sees that latter as subjectively better, with cleaner and more balanced and coherent design.

> I think it's fine for the sidebar to have a smaller font than the main text so users can clearly see the difference between the two.

A basic axiom of good design, one of the very few my mom really drilled into me, is for a clean, professional, non-distracting look focused on the content, one should minimize the number of different font sizes and styles, and if different font sizes *are* used for a good reason, they should be *enough* different that they look clear and purposeful; conversely, if they are too close together they should either be unified or made further apart since it looks sloppy, incoherent and they don't do their job distinguishing the content. 

Right now, the sidebar l1 links (Overview, Installation) and l2 links (page headings) are 16.8 px and 15.4 px, only 0.8 px (+5%) and 0.6 px (-3.75%) away from the body text size of 16 px; this is close enough for neither to be distinctly different from the body text , but far enough to look sloppy and incoherent. Therefore, I bumped the sizes down slightly, to 16px and 15px respectively, so the larger size matches the body text while the smaller is 1 px down (I didn't want to go further to avoid it becoming too small). 

Alternatively, if you think its too small, we could bump the sizes up, to e.g. 18 px and 16 px, but that would make the already large sidebar even larger. We could also do something more radical to make the sizes/styles more distinct from the body text, but I really don't think that's necessary since it is already clearly separated by the larger amount of white space you added (which will likely be increased much further with #63 ), as well as the underline style on all the text and the separate "Navigation" header.

Similarly, the line height is 1.60 (26.8 px) vs. 1.57 (25.1 px) for the body; absolutely no reason to have them different when they are that close since they don't produce a clear visual contrast. Finally, the Navigation heading (24 px) is awkwardly between the sizes of the h2 (25.6 px, +6%) and h3 (20.8 px, -13%) body headings (which are already rather close together) and a tad small besides, so I just bumped it up a tad to match the h2 size.

Incidentally, I just noticed the admonition titles looked a bit large and see that they have the exact same as the current nav header, so I'll fix that momentarily.

ccordoba12:
The sidebar format is adjusted by the theme itself, so I'd prefer to leave it like that. Obviously this is a matter of taste, as you said, but I really don't like your changes, sorry.

CAM-Gerlach:
> The sidebar format is adjusted by the theme itself

To be fair, isn't most of the other things we changed all deviations from the theme? In the [original Alabaster theme](https://alabaster.readthedocs.io/en/latest/) the sidebar main font size is the same as the text size, at least within 0.2 px; the sidebar heading size is matched to the Admonition head size, and in turn the Admonition head size is actually slightly *smaller* than the h3 size (while it is considerably bigger here, since it didn't get bumped down along with the other sans-ified headings in #61 ).

> Obviously this is a matter of taste, as you said, but I really don't like your changes, sorry.

Right, but aside from our offsetting subjective opinions as you mention, there is also the objective principles of design I've cited in favor of the changes.

CAM-Gerlach:
@ccordoba12 However, if neither of those lines of reasoning has much chance of changing your mind on this, then I don't want to waste your time arguing this any further. If that's the case, you can just go ahead and close this, and we can move on. Honestly, if I had to pick only one, #63 (which was directly inspired by your Github Books "counterexample" in another long and stubborn font size discussion) is much more important to the overall look and I'd much rather spend whatever cachet I have in advocating the timely implementation of those changes.

CAM-Gerlach:
I guess I'll just go ahead and close this, since there's been no response for a few weeks.

ccordoba12:
Instead of this, I'd prefer our sidebar to be of fixed size and completely to the left, like the one we talked about in Git books.

However, @dalthviz is quite busy with the ipdb kernel, sorry.

CAM-Gerlach:
@ccordoba12 They are two different changes doing different things for different purposes; this one tweaks the font size to match the main text in order to harmonize the appearance, while that one adjusts the sidebar and body layout and alignment in order to be more responsive to different screen sizes and avoid cutting off the text. However, as mentioned above, I agree the latter is much more important than this, relatively speaking.

PR #69, opened 2018-07-09
PR: Update installation docs to reflect need to install correct version of spyder-kernels for 3.3+

### LINKED ISSUE #55: Update installation instructions to account for new spyder-kernels dependency
## Issue Description

Per spyder-ide/spyder#7337 (and ultimately due to  spyder-ide/spyder#7306 ), installation instructions will need to be updated to clarify the need for an additional dependency, ``spyder-kernels``, which will in turn need to be for the correct version of spyder (3.x or 4/master).

Installation instructions docs, specifically installing from source and setting up a development environment sections.

Reminding the user to install the dependency, and the different procedure for ``master`` vs. ``3.x``.

### Other information

Spyder 3.3 onward, and particularly for work on ``master``.

CAM-Gerlach:
### Pull Request Checklist

## Description of Changes

Should fix the main issue behind spyder-ide/spyder#7337 Hopefully clarifies the install procedure when cloning from source under Spyder 3.3+. Let me know if I missed anything or something was inaccurate (particularly the install command using ``conda``--will it require ``-c spyder-ide``? I couldn't actually test it since 3.3 doesn't appear to be on ``conda`` yet.

### Issue(s) Resolved

Fixes #55

### DIFF
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -267,15 +267,21 @@ In summary:
 
    .. note::
 
-      Following the separation of ``spyder-kernels`` into its own modular package in Spyder 3.3.0, you'll need to have the corresponding version of that package available—``0.x`` for Spyder 3 (``3.x`` branch), and ``1.x`` for Spyder 4 (``master`` branch).
+      Following the separation in v3.3 of Spyder's console code into its own package, ``spyder-kernels``, you'll need to have the corresponding version of it available—``0.x`` for Spyder 3 (``3.x`` branch), and ``1.x`` for Spyder 4 (``master`` branch).
       The above procedure will install the ``0.x`` version; to test the ``master`` branch (Spyder 4), you'll need to install the corresponding ``1.x`` version of ``spyder-kernels``.
-      This can be done via two methods: installing the correct version via ``conda`` (or ``pip``):
+      This can be done via two methods: installing the ``1.x`` version via ``conda``:
 
       .. code-block:: bash
 
-         conda install spyder-kernels=1.*
+         conda install -c conda-forge spyder-kernels=1
 
-      (and using ``conda install spyder-kernels=0.*`` to switch back to the Spyder 3 version), or by ``clone``-ing the `spyder-kernels git repository`_ to somewhere on your path and checking out the appropriate branch (``0.x`` or ``master``) corresponding to the version of Spyder (3 or 4) you would like to run.
+      or ``pip``:
+
+      .. code-block:: bash
+
+         pip install --no-deps -U spyder-kernels==1.*
+
+      (and using the same respective command, replacing ``1`` with ``0``, to switch back to the Spyder 3 version), or by ``clone``-ing the `spyder-kernels git repository`_ to somewhere on your path and checking out the appropriate branch (``0.x`` or ``master``) corresponding to the version of Spyder (3 or 4) you would like to run, and running the commend ``pip install -e`` at the root.
       For any non-trivial development work, keeping two separate virtual environments (with ``conda-env`` or ``venv``) for Spyder 3 and 4 makes this process much quicker and less tedious.
 
 #. Install `Git`_, a powerful source control management tool.
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -264,6 +264,19 @@ In summary:
 
    This installs all of Spyder's dependencies into the environment along with the stable/packaged version of Spyder, and then removes Spyder itself.
 
+   .. note::
+
+      Following the separation of ``spyder-kernels`` into its own modular package in Spyder 3.3.0, you'll need to have the corresponding version of that package available—``0.x`` for Spyder 3 (``3.x`` branch), and ``1.x`` for Spyder 4 (``master`` branch).
+      The above procedure will install the ``0.x`` version; to test the ``master`` branch (Spyder 4), you'll need to install the corresponding ``1.x`` version of ``spyder-kernels``.
+      This can be done via two methods: installing the correct version via ``conda`` (or ``pip``):
+
+      .. code-block:: bash
+
+         conda install spyder-kernels=1.*
+
+      (and using ``conda install spyder-kernels=0.*`` to switch back to the Spyder 3 version), or by ``clone``-ing the `spyder-kernels git repository`_ to somewhere on your path and checking out the appropriate branch (``0.x`` or ``master``) corresponding to the version of Spyder (3 or 4) you would like to run.
+      For any non-trivial development work, keeping two separate virtual environments (with ``conda-env`` or ``venv``) for Spyder 3 and 4 makes this process much quicker and less tedious.
+
 #. Install `Git`_, a powerful source control management tool.
 
 #. Clone the Spyder source code repository with the command:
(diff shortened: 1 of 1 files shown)


CAM-Gerlach:
@ccordoba12

doc/installation.rst
```diff
@@ -264,6 +265,19 @@ In summary:
 
    This installs all of Spyder's dependencies into the environment along with the stable/packaged version of Spyder, and then removes Spyder itself.
 
+   .. note::
+
+      Following the separation of ``spyder-kernels`` into its own modular package in Spyder 3.3.0, you'll need to have the corresponding version of that package available—``0.x`` for Spyder 3 (``3.x`` branch), and ``1.x`` for Spyder 4 (``master`` branch).
```
ccordoba12: I think this is better:

> Following the separation of the code that powers Spyder's consoles into its own modular package (called ``spyder-kernels``) ...

doc/installation.rst
```diff
@@ -264,6 +265,19 @@ In summary:
...
+      The above procedure will install the ``0.x`` version; to test the ``master`` branch (Spyder 4), you'll need to install the corresponding ``1.x`` version of ``spyder-kernels``.
+      This can be done via two methods: installing the correct version via ``conda`` (or ``pip``):
+
+      .. code-block:: bash
+
+         conda install spyder-kernels=1.*
```
ccordoba12: This is not possible with `conda` because I instructed Anaconda guys to *not* package spyder-kernels >= 1.0.

So, this needs to be

    pip install --no-deps -U spyder-kernels

doc/installation.rst
```diff
@@ -264,6 +265,19 @@ In summary:
...
+
+      (and using ``conda install spyder-kernels=0.*`` to switch back to the Spyder 3 version), or by ``clone``-ing the `spyder-kernels git repository`_ to somewhere on your path and checking out the appropriate branch (``0.x`` or ``master``) corresponding to the version of Spyder (3 or 4) you would like to run.
```
ccordoba12: Two things here:

1. Since you need to use pip to install spyder-kernels >= 1.0, you also need to use it to install 0.x, so you need to change `conda install spyder-kernels=0.*` for `pip install spyder-kernels==0.*`.
2. You don't achieve anything by simply cloning the spyder-kernels repo and checking out the right branch. You also need to run `pip install -e .` to install it in your env (I told you that before).

doc/installation.rst
CAM-Gerlach: Thanks; I back-ported the same revised line you requested on the Contributing Guide in spyder-ide/spyder#7418 .

doc/installation.rst
CAM-Gerlach: Since the ``1.x`` version is available on ``conda-forge``, would ``conda install -c conda-forge spyder-kernels=1`` be better to suggest? That's how I have it now, and it seems to work just fine so far running from source. Is the risk of problems installing a package from a different channel with conda greater than using ``pip``? 

Seems a bit of inconvenience and potential problems for developers, but I assume its to avoid a (bigger) issue with users accidentally installing the wrong version. If using ``pip``, does the default conda-installed version also then need to be removed, in order to not have two versions?

CAM-Gerlach: Also, once this gets cleared up, the Contributing Guide version needs to be fixed too since the problem wasn't caught there.

doc/installation.rst
CAM-Gerlach: > Since you need to use pip to install spyder-kernels >= 1.0, you also need to use it to install 0.x,

I re-wrote that section to clarify the need to use the same command to switch. Also, in the initial instructions, the whole dependency stack, including spyder-kernels, gets installed via ``conda install spyder``. If ``conda install -c conda-forge spyder-kernels=1`` works in the above block as well as ``pip`` (which it seems to in my limited testing), then it avoids this potential problem.

> You also need to run ``pip install -e``

Yup, back-ported.

> I told you that before

Yes, on my PR *after* this one, spyder-ide/spyder#7418 , that had a downstream version of that text. I hadn't yet back-ported the fixes upstream to here.

doc/installation.rst
CAM-Gerlach: UPDATE: Apparently, the 1.0.1 version is on conda defaults, at least according to my ``conda info spyder-kernels`` command.

CAM-Gerlach:
@ccordoba12 Since both versions of Spyder-kernels (1.0.1 and 0.2.4) *are* in the ``main`` defaults-channel repo and the installation command ``conda install spyder-kernels=1.*`` works as originally written, can we just revert back to that, or will ``1.0.1`` be explicitly removed per your instructions? For now, I've changed the command accordingly, so if that won't change in the immediate future then this should be ready to merge on my end since I believe I addressed everything else IIRC; if not, then I can revert or we can handle it differently.

ccordoba12:
> Since both versions of Spyder-kernels (1.0.1 and 0.2.4) are in the main defaults-channel repo

That's only the case for Windows, so please revert your last changes.

CAM-Gerlach:
> That's only the case for Windows, so please revert your last changes.

Okay, thanks; dropped the commit changing it. I'll update the contributing guide version too as originally mentioned.

doc/installation.rst
```diff
@@ -264,6 +265,25 @@ In summary:
...
 
+   .. note::
+
+      Following the separation in v3.3 of Spyder's console code into its own package, ``spyder-kernels``, you'll need to have the corresponding version of it available—``0.x`` for Spyder 3 (``3.x`` branch), and ``1.x`` for Spyder 4 (``master`` branch).
+      The above procedure will install the ``0.x`` version; to test the ``master`` branch (Spyder 4), you'll need to install the corresponding ``1.x`` version of ``spyder-kernels``.
+      This can be done via two methods: installing the ``1.x`` version via ``conda``:
+
+      .. code-block:: bash
+
+         conda install -c conda-forge spyder-kernels=1
```
ccordoba12: This is not good because it'd lead to mix packages from conda-forge and defaults. Please change it to

    conda install -c spyder-ide spyder-kernels=1

I'll move the spyder-kernels package to our own channel a soon as I can.

doc/installation.rst
```diff
@@ -264,6 +265,25 @@ In summary:
...
+
+      or ``pip``:
+
+      .. code-block:: bash
+
+         pip install --no-deps -U spyder-kernels==1.*
```
ccordoba12: Since we're going to use conda to install `spyder-kernels`, this could be simpler:

    pip install spyder-kernels==1.*

doc/installation.rst
CAM-Gerlach: Okay, sure—that does sound better. I tried that first but it didn't work at least back when I tested it since presumably the packages weren't posted there yet as you mention. Fixed.

doc/installation.rst
CAM-Gerlach: Okay, gotcha. I figured that might be but wasn't 100% sure. Done.

CAM-Gerlach:
@ccordoba12 Thanks, changed as requested.

ccordoba12:
@CAM-Gerlach, let us move the spyder-kernels packages to our channel before merging this one.

@dalthviz, what's the command to move packages from conda-forge to our channel?

dalthviz:
I think it should be something like 
```
anaconda copy conda-forge/spyder/3.3.0 --to-owner spyder-ide
```
[Here some docs](https://docs.anaconda.com/anaconda-repository/user-guide/tasks/pkgs/copy-pkg)

ccordoba12:
Thanks a lot @dalthviz! The packages are in our channel now.

PR #84, opened 2018-09-20
PR: Add mention of kernel ID number in description of connect to kernel dialog

CAM-Gerlach:
### Pull Request Checklist

* [x] Read and followed this repo's [Contributing Guidelines](https://github.com/spyder-ide/spyder/blob/master/CONTRIBUTING.md)
* [x] Based your PR on the latest version of the correct branch (master or 3.x)
* [x] Checked your writing carefully for correct English spelling, grammar, etc
* [x] Described your changes and the motivation for them below

## Description of Changes

As suggested by @bcolsen in spyder-ide/spyder#7914 , I've added a mention of the use of the kernel ID in the connect to kernel dialog for the connection field. Also did a bit of minor cleanup/consistency conformance to the line spacing in that lists, to make it easier to follow.

PR #85, opened 2018-09-20
PR: Be clearer and more consistent about runtime dir placeholder name

CAM-Gerlach:
### Pull Request Checklist

## Description of Changes

As suggested by @bcolsen in a followup to #84 , I've made the Jupyter runtime dir placeholder more repristentative and consistant, and the example PID number more generic and matching the dialog in the Connect the Kernel procedure.

### DIFF
--- a/doc/ipythonconsole.rst
+++ b/doc/ipythonconsole.rst
@@ -43,16 +43,16 @@ To connect to an external kernel,
    If using Spyder 3.3.0 or later, you'll need to do so with ``python -m spyder_kernels.console`` (after you've first installed ``spyder-kernels`` on the host with ``<conda/pip> install spyder-kernels``).
    If using a version of Spyder before 3.3.0, ``ipython kernel`` should work to launch the kernel, albeit without certain Spyder-specific features.
 
-#. Copy the connection file (:file:`{runtime-dir}/kernel-{pid}.json`) to the machine you're running Spyder on (if remote) or note its location (if local).
+#. Copy the connection file (:file:`{jupyter/runtime/dir/path}/kernel-{pid}.json`) to the machine you're running Spyder on (if remote) or note its location (if local).
 
-   You can get the :file:`{runtime-dir}` path by executing ``jupyter --runtime-dir`` in the same Python environment as the kernel.
+   You can get :file:`{jupyter/runtime/dir/path}` by executing ``jupyter --runtime-dir`` in the same Python environment as the kernel.
 
 #. Click :guilabel:`Connect to an existing kernel` from the :guilabel:`Console` menu or the :guilabel:`IPython Console` pane's "Gear" menu.
 
 #. Browse for or enter the path to the connection file from the previous step.
    If you're connecting to a local kernel, click :guilabel:`Ok` and Spyder should connect to the kernel; if a remote kernel, proceed to the final step.
 
-   As a convenience, kernel ID numbers (e.g. ``20345``) entered in the connection file path field will be expanded to :file:`{jupter_runtime_dir}/kernal-{id}.json` on your local machine.
+   As a convenience, kernel ID numbers (e.g. ``1234``) entered in the connection file path field will be expanded to :file:`{jupyter/runtime/dir/path}/kernal-{id}.json` on your local machine.
 
 #. If connecting to a remote kernel over ``ssh``, check the appropriate box and type the full hostname you're connecting to (in the form :file:`{username}@{hostname}:{port-number}`).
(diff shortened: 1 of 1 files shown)


doc/ipythonconsole.rst
```diff
@@ -43,16 +43,16 @@ To connect to an external kernel,
    If using Spyder 3.3.0 or later, you'll need to do so with ``python -m spyder_kernels.console`` (after you've first installed ``spyder-kernels`` on the host with ``<conda/pip> install spyder-kernels``).
    If using a version of Spyder before 3.3.0, ``ipython kernel`` should work to launch the kernel, albeit without certain Spyder-specific features.
 
-#. Copy the connection file (:file:`{runtime-dir}/kernel-{pid}.json`) to the machine you're running Spyder on (if remote) or note its location (if local).
+#. Copy the connection file (:file:`{jupter/runtime/dir}/kernel-{pid}.json`) to the machine you're running Spyder on (if remote) or note its location (if local).
```
ccordoba12: jupter -> jupyter

ccordoba12: The same error is present in your other changes.

CAM-Gerlach: Well, that's awfully embarrassing. So much for

> [x] Checked your writing carefully for correct English spelling, grammar, etc

Thanks for the catch; fixed.

CAM-Gerlach:
@ccordoba12 That's basically how I had it before, ``{jupyter_runtime_dir}`` (obviously, we don't want spaces in an example path). However, I added the path deliminators to make it more clear that ``{jupyter/runtime/dir}`` was a path and not just a directory name, as it otherwise implies. We could also make it ``jupyter_runtime_dir_path``, but that makes it even longer and requires reading comprehension, vs. simply *showing* it is a path directly by our example. Conversely, I'm not sure I understand what clarity advantages exist that are cite above as justification for removing them.

CAM-Gerlach:
@bcolsen Any thoughts?

CAM-Gerlach:
> That people is used to see paths on Windows with `\` instead of `/`.

Any a remotely proficient computer user regardless of platform is almost certain to be familiar with ``/`` as a path separator, much less a developer using open-source software:

* If they've ever used the web or seen a URL they are well familiar with ``/`` as the path deliminator
* Any other OS they've used would have used ``/`` 
* Windows fully accepts both, so using it instead of ``\`` works just as well anyway
* Many common applications and programming languages use ``/`` even on Windows (e.g. Git, many open-source applications, R/Rstudio, the other common language and IDE for data science, etc.);
* In Python strings that represent paths, ``/`` is often used even on Windows, and certainly all cross-platform applications, since it works across all common platforms and avoids double-escapes or raw stirngs
* It is just basic computer knowledge that even a beginning programmer is likely to know
* For what it's worth, I'm still a n00b and a Windows user, and I've known about ``/`` as the path deliminator since I was a child

Furthermore, paths are less avoidably used a number of other places in the documentation, and of course ``/`` is used there as well; this seems to be the universal convention for other open-source documentation (that isn't exclusively focused on Windows, at least), so this just follows the established convention.

Nevertheless, I can add ``/path`` to the end to make it more explicit it is a path; I think that's a good idea regardless and it will eliminate any ambiguity as to those being path deliminators.

CAM-Gerlach:
Oh, and there's also the fact that there's a ``/`` as the path separator literally one character after the part you copied anyway in 2 of the three places its used jejeje...

```rst
:file:`{jupyter/runtime/dir}/kernel-{pid}.json`
```

bcolsen:
I have seen the `/` used in tutorials and documentation as a `path/to/your/file` kind of way and the `/` operator is used to concatenate python path objects so paths in python are `/` centric, like most OS's(just not the most popular one).

CAM-Gerlach:
> OS's(just not the most popular one).

Somewhat tangential, but to note, the ratio is much closer in terms of Spyder users (and programmers/scientists/engineers/data analysts in general) vs. the wider desktop computing world, at least judging from some Quick Maths™ in terms of ``Operating System: `` tags in issues reported—the ratio is something between 60%/30%/10% and 45%/40%/15% Win/Linux/Mac depending on exactly how things are double-counted between different search queries, so Windows users may be in the plurality, but perhaps not even in the majority of the people using Spyder.

ccordoba12:
> There's also the fact that there's a / as the path separator one character after the part you copied anyway in 2 of the three places its used jejeje xD

Jaja, ok, no problem. Thanks @CAM-Gerlach and @bcolsen for the discussion.

PR #116, opened 2020-06-04
PR: Build api docs with sphinx

goanpeca:
This enables building ALL Spyder modules docs with sphinx-autodoc, assuming that you make a clone of spyder and place it in a folder called `spyder-repo, so

```
spyder-docs/
  doc
  ...
  spyder-repo/
    bootstrap.py
    ...
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,10 @@ spyder_crash.log
 *.tmp
 *.temp
 *.bak
+
+spyder/
+spyder-repo/
+
+doc/modules.rst
+doc/spyder*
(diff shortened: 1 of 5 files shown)


goanpeca:
@CAM-Gerlach this is finally building! :-) 

No need to merge yet, but at least things are building :-) 

It seems like we will need a 4.x branch so we can separate this dev docs that will be aimed at Spyder 5.

Thoughts?

goanpeca:
Not really part of the API work. Closing.

PR #127, opened 2020-07-03
PR: Use new theme

dalthviz:
A preview:

docs


ccordoba12:
@dalthviz. I think we should probably hide the `Open Chat` button when the screen is small because it takes too much space. And we have to take into account that in Spyder 5 our plan is to embed our docs in the Help pane.

@juanis2112, what do you think?

juanis2112:
I wouldn't suggest removing it, but making the red square much smaller in the lower right corner of the whole page (so it doesn't actually move when scrolling down, but you only see it when you go to the bottom) and instead of having the full text just an icon that indicates that it's the chat button. Something like this:
Screen Shot 2020-07-05 at 10 21 02 PM

(But in white so it has enough contrast with the red square)

@ccordoba12 what do you think?

ccordoba12:
Very nice suggestion @juanis2112!

But what about adding that button next to the magnifying glass and outline icon at the bottom? I mean, instead of showing it in the right lower corner. That way it won't take more valuable screen space.

dalthviz:
Note: Maybe I should add the support for the Gitter chat inside the theme

ccordoba12:
I think that's a great idea @dalthviz!

CAM-Gerlach:
Hey @dalthviz , even if its not desired that I review the content of the theme itself at this time, I do need to understand the details of how this changes the build system, and I have a few questions about the specific changes, being I'm the person responsible for maintaining it. I'll leave them as review comments.

Could you include at least a high-level summary in the OP about how this theme is implemented, how it modifies the doc build process, how changes to the theme are picked up by the docs site and what's the deal with Windows? Thanks!

.travis.yml
```diff
@@ -4,24 +4,22 @@ language: python
 dist: bionic
 
 python:
-  - 3.7
+  - 3.6
```
CAM-Gerlach, line 7: Can you explain the reason this was downgraded? Key dependencies not being compatible with Python 3.7, which has been out for over two years, is a significant worry given that Python 3.6 only has another year and a half to live.

.travis.yml
```diff
@@ -4,24 +4,22 @@ language: python
...
+    - CI="true"
+    - secure: "Kyk5NrwAnxbQoanhqkI/XGMXLzJKaxUiN3j0w1j63r1qJUAntKf9lY7iRupmdc6ZxgoJhtNYFAw+wVGaWQ6aoqWLZOGJ03GX+UVfWNtB+kKhsSWzVxdjxoup/jsRJquIZ2TxSKTNcXSrqIPMQbtTPDFWkmpn8pZR7ply3OBFRQY2ZVAIVkE2pl16rNw/vChfs3QA/duNWFwhRVNr/O5u/Af0Lmm7gjPPfjZOObg/hPLklH7KOMSf0Br0pwBatG1Ld60DQp49piMSsEty0xlGVBs1Dh5y8hd4HIzWOPW9cdd4klSljH37sxyInlOC2crmC3ey64oPwMqgdfGRDz/CCnTOwW0J/UTJLekNx9cN8Sp2RfKk1qcLaVuODUJiKzFQPZw6dC+cypyVtTS6gmS1D1yDaYXNdtTtYjZc9IY4MH6Z2alzJLq0WLyJJcyyG6bG449KVC523kJY/r/klVwq5zyyaP/pXol0zc8rVh6jN9XLJb8N308ZvN6/5bTIY0qUqICT2vZEmPwjpIpJCrpnrLQ6TQWCsk+htEAYHXNUKm1bA+I1gpCGQVfK6YUGF9SERnao71H2/nB9SPgAdK+TmMGApUAHjQGL6e46SQhHC7Ct+TE52nFlM/tzHxMsPFdgyqNKJVsNwwtMcaA5YvpAan+cDi9h++3+CO3hV4ysUgI="
 
 install:
   - pip install -U pip
   - pip install sphinx doctr
+  - pip install git+https://github.com/spyder-ide/spyder-docs-sphinx-theme.git@develop_spyder
+  - make docs
```
CAM-Gerlach, line 19: Is there a reason building the docs, the central component of the workflow, is set as an install step rather than a core build or script step? This seems semantically incongruous, but I'm assuming there is a technical reason for it?

CONTRIBUTING.md
```diff
@@ -79,19 +79,21 @@ conda install sphinx
 Or, if using ``pip``, you can grab it with
 
 ```bash
+python -m pip install git+https://github.com/spyder-ide/spyder-docs-sphinx-theme.git
```
CAM-Gerlach, line 82: This should be repeated under the `conda` section.

Furthermore, it should presumably follow `pip install sphinx` given Sphinx is an implicit or explicit dependency of the theme.

Is there a reason `python -m pip` is used for this step, yet `pip` is used for the next? We should be consistent and use one or the other.

CONTRIBUTING.md
```diff
@@ -79,19 +79,21 @@ conda install sphinx
...
 pip install sphinx
 ```
 
-Now, ``cd`` into the ``doc`` directory in your clone of this repo, and run
```
CAM-Gerlach, line 85: Why was this removed? The code block should have at least some description, e.g.

```markdown
Now, run the following on macOS/Linux:
```

CONTRIBUTING.md
```diff
@@ -79,19 +79,21 @@ conda install sphinx
...
-
 ```bash
-make html
+make docs
+make serve
 ```
 
-or, if on Windows
+or, if on Windows:
```
CAM-Gerlach, line 91: A colon is not used above, yet it is here

CONTRIBUTING.md
```diff
@@ -79,19 +79,21 @@ conda install sphinx
...
 
 ```cmd
+cd doc
 make.bat html
```
CAM-Gerlach, line 95: So does this new infra not work at all on Windows, works similarly, or somewhere in between? Could you please document this?

Makefile
```diff
@@ -0,0 +1,48 @@
...
+docs: ## generate Sphinx HTML documentation, including API docs
+	$(MAKE) -C doc clean
+	$(MAKE) -C doc html
+	# $(MAKE) -C doc linkcheck
+
+servedocs: doc ## compile the docs watching for changes
+	watchmedo shell-command -p '*.rst' -c '$(MAKE) -C doc html' -R -D .
+
+install: clean ## install the package to the active Python's site-packages
+	python setup.py install
```
CAM-Gerlach, line 45: As `pip install .` is normally always preferable to `python setup.py install` as it as a safer, more robust, strict superset thereof, is a reason it isn't used here?

Makefile
```diff
@@ -0,0 +1,48 @@
...
+
+autodocs: ## generate Sphinx HTML documentation, including API docs
+	rm -f doc/spyder.*
+	rm -f doc/modules.rst
+	sphinx-apidoc -o doc/ spyder-repo/spyder/ *tests*
+
+docs: ## generate Sphinx HTML documentation, including API docs
+	$(MAKE) -C doc clean
+	$(MAKE) -C doc html
+	# $(MAKE) -C doc linkcheck
```
CAM-Gerlach, line 39: Why was this disabled? Without it, links are likely to go stale and die.

doc/conf.py
```diff
@@ -35,37 +41,37 @@
...
 # The master toctree document.
-master_doc = 'index'
+master_doc = "index"
 
 # General information about the project.
-project = 'Spyder'
-copyright = " 2018 Spyder Doc Contributors <a href='https://opensource.org/licenses/MIT' target='_blank'>MIT License</a>"
-author = 'The Spyder Doc Contributors'
+project = "Spyder"
+copyright = " 2020 Spyder Doc Contributors "
```
CAM-Gerlach, line 64: Do we have an easy way to embed a link to a license and/or legal notices in the footer (or somewhere else on all pages)? We're going to need it.

doc/conf.py
```diff
@@ -76,109 +82,132 @@
...
-    'github_banner': False,
-    'travis_button': False,
-    'codecov_button': False,
-    'extra_nav_links': {
-        'Troubleshooting': ('https://github.com/spyder-ide/spyder/wiki/'
-                            + 'Troubleshooting-Guide-and-FAQ'),
-        'Spyder Wiki': 'https://github.com/spyder-ide/spyder/wiki',
+# CI = True
+# TRAVIS_BRANCH = 'master'
+if True or CI and TRAVIS_BRANCH == 'master':
```
CAM-Gerlach, line 129: This is always going to be `True`...?

dalthviz:
Hi @CAM-Gerlach thanks for the feedback, actually much of the changes in the script were proposed initially by @goanpeca but sure I will check (also, take into account some of the changes (like the `True`) are there to be able to make a local build to create the preview of how things are looking but will be removed or changed). As soon as this PR gets to a stable state I will let you know for sure :+1:

CAM-Gerlach:
Okay, thanks @dalthviz ! Didn't know how soon you all were planning to merge this so I wanted to make sure I got my questions in before then. Yeah, just let me know whenever its stable. Thanks!

doc/conf.py
dalthviz: Nop

doc/conf.py
dalthviz: Yep

Makefile
dalthviz: No idea, I re-added it

Makefile
dalthviz: I don't even now if this is necessary. Will remove it for now

CONTRIBUTING.md
dalthviz: Sure :+1:

.travis.yml
dalthviz: No idea, will check with 3.7

.travis.yml
dalthviz: Not that I'm aware, will change.

dalthviz:
@CAM-Gerlach this is ready for a review :)

.travis.yml
CAM-Gerlach: Thanks!

.travis.yml
CAM-Gerlach: Thanks!

CONTRIBUTING.md
CAM-Gerlach: All looks good, thanks @dalthviz !

CONTRIBUTING.md
CAM-Gerlach: Great!

CONTRIBUTING.md
CAM-Gerlach: Thanks!

Makefile
CAM-Gerlach: Okay, thanks!

Makefile
CAM-Gerlach: Possibly for local testing since there were some broken links I've since fixed. Thanks!

doc/conf.py
CAM-Gerlach: Thanks! I'll update it further after the theme is merged and stable.

doc/conf.py
CAM-Gerlach: Thanks!

PR #131, opened 2020-07-07
PR: Add message indicating that documentation is a WIP

dalthviz:
Add WIP message:

image

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -138,7 +138,7 @@ if CI and TRAVIS_BRANCH == 'master':
         ],
         "use_edit_page_button": True,
         "gitter_room": "spyder-ide/public",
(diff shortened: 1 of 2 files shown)


doc/conf.py
```diff
@@ -137,7 +137,8 @@
             }
         ],
         "use_edit_page_button": True,
-        "gitter_room": "spyder-ide/public"
+        "gitter_room": "spyder-ide/public",
+        "alert_message": "This documentation is still a work in progress!"
```
juanis2112, line 141:
```suggestion
        "alert_message": "This documentation is still a work in progress. Check out the [Spyder 3 docs](https://docs.spyder-ide.org) for the stable version"
```

I suggest this message instead so people can access to our current docs.

PR #143, opened 2020-07-31
PR: Add a custom bottom margin to sections with cards

ccordoba12:
This is necessary to have more separation between the cards of that section and the header of "Join our community".

## Before

Selección_025

## After

Selección_027

### DIFF
--- a/doc/_static/custom_styles.css
+++ b/doc/_static/custom_styles.css
@@ -16,8 +16,8 @@ div.fasb h3 {
 }
 
 /* Custom bottom margin for the Where to go now section */
-#where-to-go-now {
-	margin-bottom: 18px;
+.section .sphinx-bs {
+  margin-bottom: 18px;
 }
 
 /* Custom color for FA icons in the Where to go now section */
@@ -26,7 +26,8 @@ div.fasb h3 {
   width: 32px;
 }
 
-/* Custom colors for specific FA icons in the index page */
+/*** Custom colors for specific FA icons in the index page ***/
+
 div.fa-book::before {
   color: #6a3514;
 }
--- a/doc/_static/custom_styles.css
+++ b/doc/_static/custom_styles.css
@@ -15,6 +15,11 @@ div.fasb h3 {
   display: inline-block;
 }
 
+/* Custom bottom margin for the Where to go now section */
+#where-to-go-now {
+	margin-bottom: 18px;
+}
+
 /* Custom color for FA icons in the Where to go now section */
 #where-to-go-now div.fasb::before {
   color: #36557c;
@@ -22,7 +27,6 @@ div.fasb h3 {
 }
 
 /* Custom colors for specific FA icons in the index page */
-
 div.fa-book::before {
   color: #6a3514;
 }


ccordoba12:
@juanis2112, please take a look at this one.

doc/_static/custom_styles.css
```diff
@@ -15,14 +15,18 @@ div.fasb h3 {
   display: inline-block;
 }
 
+/* Custom bottom margin for the Where to go now section */
+#where-to-go-now {
+	margin-bottom: 18px;
+}
+
```
CAM-Gerlach, line 22:
```suggestion
/* Standard bottom margin for the card sections */
.section .sphinx-bs {
  margin-bottom: 30px;
}

```

Instead of tying the rule to this one particular section title, it makes more sense and is less fragile and bespoke if we instead actually solve the root problem, the cards element not having enough bottom margin, unlike the other elements.

IO we were to take a bespoke approach, it should be simply adding a line break (`|`) in the rst source file, which won't break if the section name changes and add another whole custom CSS selector.

doc/_static/custom_styles.css
```diff
@@ -15,14 +15,18 @@ div.fasb h3 {
...
 /* Custom color for FA icons in the Where to go now section */
 #where-to-go-now div.fasb::before {
   color: #36557c;
   width: 32px;
 }
 
 /* Custom colors for specific FA icons in the index page */
-
 div.fa-book::before {
```
CAM-Gerlach, line 30:
```suggestion

div.fa-book::before {
```

This blank line  is here because the comment describes all the selectors beneath it, not just the one immediately below as would be implied by a lack of one.

doc/_static/custom_styles.css
CAM-Gerlach: Also, fixed a hard tab in the file, inconsistent with the rest of the formatting.

ccordoba12: Why 30px? That's too really much space.

doc/_static/custom_styles.css
ccordoba12: It looks odd with respect to the other comments. Please suggest one that applies specifically to the div below then.

doc/_static/custom_styles.css
CAM-Gerlach: Its the standard margin everything else uses, plus the existing padding on the element. I don't think its excessive, but if you'd prefer it with just the standard margin and not the extra padding, the following will accomplish that:

```suggestion
/* Standard bottom margin for the card sections */
.section .sphinx-bs {
  margin-bottom: 30px;
  padding-bottom: 0 !important;
}

```

doc/_static/custom_styles.css
CAM-Gerlach: Yeah, but then every icon coloring would need a comment, which would get repetitive fast and take up space without really explaining anything that isn't obvious from the first comment. Instead, we can set the comment off as a heading for multiple selectors via formatting like this, so its clear its not supposed to parallel with the others:

`/*** Custom colors for specific FA icons in the index page ***/`

doc/_static/custom_styles.css
ccordoba12: Sure, but please take a look at the screenshots I added in the description. I carefully selected 18px because that's the amount that looks about the same between the paragraph before "Where to go " and that header, and between the cards and "Join our community".

I don't think this needs to be about using the same amount as everywhere else, just making things look uniform.

doc/_static/custom_styles.css
ccordoba12: Please don't plan for a non-existing future. If more comments are necessary, we'll see what to do then. For now, please suggest a comment that makes sense for the adjustments we have below.

CAM-Gerlach: I'm not referring to a "non-existing [sic] future", I'm talking about right now. As I've stated previously, the CSS comment describes all three blank-line-separated selector blocks following it, not just the one it is immediately below it as the lack of a blank line below it implies, following the [Google CSS style guide](https://google.github.io/styleguide/htmlcssguide.html) and standard practice (including in the Pandas theme ours is based on).

I've suggested an alternative that tries to address your personal aesthetic opinion about it not "looking odd with respect to the other comments", while keeping this conceptual clarity intact, but if you still don't like the look of it, we can follow the style used by our theme and make _all_ comments section comments with blank lines before and after and not use single-block comments at all, so there is no visual inconsistency between any of them, like this:

<details>

```
/* Ensure FA icons applied to section divs display inline */

div.fasb h3 {
  display: inline-block;
}

/* Custom colors for specific FA icons in the index page */

div.fa-book::before {
  color: #6a3514;
}

div.fa-globe-americas::before {
  color: #1c4489;
}

div.fa-user-friends::before {
  color: #f6951f;
}
```

</details>

doc/_static/custom_styles.css
CAM-Gerlach: Okay, I understand your point. I've modified the below to add the extra 4 px to be identical in appearance to your original (1 em (16 px) padding + 18 px margin = 34 px), just organized in a more coherent manner. As I mentioned before, I agree extra padding is justifiable to account for the shadow falloff, which makes the element appear larger than its nominal bounding box, and I should have just used values that would be precisely equivalent to what you had originally—my mistake.

```suggestion
/* Standard bottom margin for the card sections */
.section .sphinx-bs {
  margin-bottom: 30px;
  padding-bottom: 4px !important;
}

```

ccordoba12: > Okay, thanks. I've modified the below to add the extra 4 px to be identical in appearance to your original (1 em (16 px) padding + 18 px margin = 34 px)

@CAM-Gerlach, this is your way of doing this change, which is exactly like mine (as you're clearly saying above). Therefore, I'm only going to take your suggestion of using `.section .sphinx-bs` so things don't depend on id of this section.

doc/_static/custom_styles.css
ccordoba12: > /*** Custom colors for specific FA icons in the index page ***/

Changing the comment to that one makes sense to me. Thanks for the suggestion.

ccordoba12:
Merging because I don't have time to keep discussing about this, being it an extremely simple change.

doc/_static/custom_styles.css
CAM-Gerlach: Sure. I can make note of this in the style guide for the future; I need to update it with a few other things we've agreed on but not formally documented.

doc/_static/custom_styles.css
CAM-Gerlach: The rendered result is the same, but the implementation is more logical, because it applies a consistent, standard margin, plus the extra padding necessary for the particular element, rather than simply applying a bespoke margin on top of a bespoke padding.

CAM-Gerlach: The commit also didn't pick up my changes to the comment so its not out of sync with the selector. I'll fix that part in #148  .

ccordoba12: Yeah, I missed the change to the comment, sorry.

PR #147, opened 2020-08-01
PR: Center images by default

ccordoba12:
When checking how the new page for Find looked, I noticed all its images were not centered. Then I went to other pages and noticed the same.

This also removes the center directive everywhere because it's not necessary anymore.


ccordoba12:
@juanis2112, please review this one.

ccordoba12:
Good idea! But that's not proper rst, so it could take people maintaining our docs in the future by surprise because it's happening behind the scenes.

So I think it's better to follow rst syntax instead of injecting that css in all pages.

CAM-Gerlach:
> However, that's not proper rst, so it could take people maintaining our docs in the future by surprise because it's happening behind the scenes.

Actually, this doesn't change anything with reST; as I explained, I was careful to write and test the selector to specifically exclude any image that has classes applied to it (which is what reST does when you add the `:align:` parameter), so `:align:` works exactly as it does otherwise. All it does is change the _default_ alignment for images without an `:align:` parameter, which is not defined in the reST specification and is the responsibility of the theme and the CSS, not reST; furthermore, even with the parameter it is entirely up to the CSS how this is actually rendered, as the [spec](https://docutils.sourceforge.io/docs/ref/rst/directives.html#images) makes clear:

> The specific behavior depends upon the browser or rendering software used.

ccordoba12:
Ok, on a second thought I think this makes sense. I'll implement it.

PR #149, opened 2020-08-02
PR: Decrease length of headers in installation guide

ccordoba12:
Current headers are way too long, which makes the outline on the right to overflow the page and be very hard to read. So I propose to use shorter headers everywhere to fix that.

## Before

Selección_032

## After

Selección_031

### DIFF
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -38,7 +38,7 @@ Using Anaconda
 
 .. important::
 
-   If in doubt, you should install Spyder via this method; it generally has the least likelihood of potential pitfalls for non-experts, and we may be able to provide limited assistance if you do run into trouble.
+   If in doubt, you should install Spyder via this method; it generally has the least likelihood of potential pitfalls for non-experts, and we will be able to provide assistance if you do run into trouble.
 
 Spyder is included by default in the `Anaconda`_ Python distribution, which comes with everything you need to get started in an all-in-one package.
 
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -15,7 +15,7 @@ Thanks!
 
 
 =================
-Try Spyder Online
+Try Spyder online
 =================
 
 Want to try out Spyder without installing it?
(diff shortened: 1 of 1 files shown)


doc/installation.rst
```diff
@@ -32,9 +32,9 @@ Just visit the `Spyder page on Binder`_ to get started using Spyder now.
 
 
 
-======================================
-Installing with Anaconda (recommended)
-======================================
+==============
+Using Anaconda
```
ccordoba12, line 36: Now I think it's necessary to add a note below this header to clarify that this is our recommended method.

CAM-Gerlach: Added, thanks.

ccordoba12: Thanks @CAM-Gerlach!

doc/installation.rst
```diff
@@ -32,20 +32,23 @@ Just visit the `Spyder page on Binder`_ to get started using Spyder now.
...
+==============
+
+.. important::
+
+   If in doubt, you should install Spyder via this method; it generally has the least likelihood of potential pitfalls for non-experts, and we may be able to provide limited assistance if you do run into trouble.
```
ccordoba12, line 41:
```suggestion
   If in doubt, you should install Spyder via this method; it generally has the least likelihood of potential pitfalls for non-experts, and we may provide assistance if you do run into trouble.
```

CAM-Gerlach:
```suggestion
   If in doubt, you should install Spyder via this method; it generally has the least likelihood of potential pitfalls for non-experts, and we may be able to provide assistance if you do run into trouble.
```

Good point, but it still needs the "be able to" to sound right.

ccordoba12: Ok, could we rephrase it a bit then? We're giving a ton of assistance right now, but (as a non-native speaker), this sounds to me "we could help you", whereas I'd like to send a message closer to "if you install Anaconda, we'll definitely help you".

CAM-Gerlach: Ah, I see. Its not just you, "we could help you" is the in fact the precise message it was originally intended to send; in my experience with user tech support I've found its almost always better to under- than over-promise, exceed people's expectations rather than risk disappointing them, and let our actions do the talking. But if we really want to promise that, what we could say is just

```suggestion
   If in doubt, you should install Spyder via this method; it generally has the least likelihood of potential pitfalls for non-experts, and we will be able to provide assistance if you do run into trouble.
```

ccordoba12: Great! I agree with this message now!

Thanks a lot @CAM-Gerlach for your patience with my requests.

CAM-Gerlach: Same to you for your patience with me :smile:

PR #176, opened 2020-08-15
PR: Fix margins for iframes

ccordoba12:
This centers our iframes (i.e. the frames that contain our Youtube videos) and fixes their bottom margin to be the same as the one for our images.

### DIFF
--- a/doc/_static/custom_styles.css
+++ b/doc/_static/custom_styles.css
@@ -50,6 +50,14 @@ div.fasb h3, div.fabb h3 {
   padding-bottom: 0.3em;
 }
 
+/* Center iframes and adjust their bottom margins */
+iframe {
+  display: block;
+  margin-left: auto;
+  margin-right: auto;
+  margin-bottom: 1.5rem;
+}
+
 /*** Custom colors for specific FA icons in the index page ***/
 
 div.fa-laptop-code::before {
--- a/doc/_static/custom_styles.css
+++ b/doc/_static/custom_styles.css
@@ -56,6 +56,7 @@ iframe {
   margin-left: auto;
   margin-right: auto;
   margin-bottom: 1.5rem;
+  margin-top: 1.0rem;
 }
 
 /*** Custom colors for specific FA icons in the index page ***/
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -369,7 +369,7 @@ class IframeVideo(Directive):
 
 
 class Youtube(IframeVideo):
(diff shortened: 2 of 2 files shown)


ccordoba12:
@juanis2112, @CAM-Gerlach, please take a look at this one and let me know what you think.

doc/_static/custom_styles.css
```diff
@@ -39,6 +39,14 @@ div.fasb h3 {
   padding-bottom: 0.3em;
 }
 
+/* Center iframes and adjust their bottom margins */
+iframe {
+  display: block;
+  margin-left: auto;
+  margin-right: auto;
+  margin-bottom: 1.5rem;
+}
```
CAM-Gerlach, line 48:
```suggestion
  margin-top: -0.7rem;
}
```

Add negative top margin to reduce height above videos

juanis2112: Agreed

juanis2112:
Thanks @ccordoba12  for fixing this. I think the centering and the bottom margin looks fine. However I agree with @CAM-Gerlach suggestion, we should reduce the top margin of the videos.

doc/_static/custom_styles.css
ccordoba12: I was thinking to do this but I decided not to because the current `margin-top` for iframes is the same as the one for images.

So, should we use this  `margin-top` for images too?

juanis2112: I guess yes, but I think that should be in another PR. For now we can fix the one in the video frames. What do you think @CAM-Gerlach ?

ccordoba12: I think I found the issue: it turns out we're using our main website's css here, which is causing a lot of other problems.

So, for now I'd suggest to merge this one as it is. Then @dalthviz or me can fix that other problem in a different PR.

CAM-Gerlach: Agreed, that makes sense. That would explain a lot of the issues we've been seeing with margins on a lot of elements.

ccordoba12: After the latest changes in the theme, we really need to increase `margin-top` instead of decrease it. @CAM-Gerlach and I think that  `1.0rem` looks good.

@juanis2112, please see how things look on Netlify and let us know what you think.

CAM-Gerlach:
@ccordoba12 I already fixed the config option in another PR so this one's going to conflict, can you drop the last two commits, rebase, force push and apply the suggestion so we can merge this as a starting point? Or if you want, I can do it for you, just lmk.

Issue #179, opened 2020-08-19
Set up UI, redirects and deployment for use with multiple versions of the docs


CAM-Gerlach:
## Issue Description

As part of #162 , we need to implement version switching in the docs. What we want is:

* [ ] Each set of docs (3.x, 4.x, and eventually 5.x) deployed to a subdirectory corresponding to the version number (which is trivial for me to do with what I've already set up with doctor)
* [ ] A dropdown menu at the top of the page to select the version of the current page of the docs to show, like the Python.org docs have (I imagine that's pretty doable for you on the theme level, e.g. in the page template generate the dropdown menu items/links using the page URL + a list of versions/subdomains in the conf.py)
* [ ] Redirects from the base doc URLs (with no version number in the URL path) to whatever version is the current stable, also like the Python.org docs do (which I'm really not sure is possible given its a static site on a host we don't control). If this isn't possible, I guess what we could do is simply have our CI script deploy whatever version is the current stable twice, once to the subdirectory and once to the root (assuming that doctr doesn't delete files and directories that don't match what's in the deploy; if it does, we could either do some hackery or just deploy the stable version to e.g. `stable`, create some kind of landing page and have to update all our links one more time (i.e. what I did in #39 )

@dalthviz says: 

Maybe is worthy to create [an issue] in the docs repo and add there any resource or new info that could help us to implement this

Checking I found [this extension](https://holzhaus.github.io/sphinx-multiversion/master/index.html
) that generates/handles the different version of docs taking into account branches/tags.

dalthviz:
Note: Possible extension to do the redirects - https://github.com/sphinx-contrib/redirects

CAM-Gerlach:
Sorry, thought I replied to this before. I'm not sure how sphinx-redirect will work at least with sphinx-multiversion, since it only redirects things relative to the Sphinx source dir, not the actual deployed dir, so I'm not sure how it would work between deployed branches. Also, if we used it, we'd have to still write code to manually build a list of all the docs to actually make the manual redirects, so unless another tool exists (which I haven't found) we will either need to write our own to do it after the build runs, or just have only the root path redirect and update any external links to point to a stable tag (which we could include a few git commands in our Travis build script to automatically create and point to the tip of the latest stable branch, so we don't have to update it manually with every PR).

 Thoughts?

CAM-Gerlach:
Sure, sounds fine to me.

[...]

PR #181, opened 2020-08-20
PR: Add transcripts to videos

juanis2112:
### Pull Request Checklist

* [ ] Read and followed this repo's [Contributing Guidelines](https://github.com/spyder-ide/spyder/blob/master/CONTRIBUTING.md)
* [ ] Based your PR on the latest version of the correct branch (master or 3.x)
* [ ] Checked your writing carefully for correct English spelling, grammar, etc
* [ ] Described your changes and the motivation for them below
* [ ] Noted what issue(s) this pull request resolves, creating one if needed

## Description of Changes

In this PR I will add all the transcripts to the videos as an scrollable element under a dropdown. 

The first 2 commits have the directives necessary to do this along with the first script but once @CAM-Gerlach is done reviewing and formatting all of them I will add the rest. 

Let me know what you think about the design. I think I'll try to talk to Isabella for some advice on colors and format. 

### Issue(s) Resolved

Fixes #181

PR #185, opened 2020-08-25
PR: Add support for docs versions

dalthviz:
### Pull Request Checklist

## Description of Changes

Currently only builds local branches that follow the regex in the config (branches with `<number>.<anything>` names and `master`).

**NOTE:** The theme being installed needs to be changed for the proper repo/branch after spyder-ide/spyder-docs-sphinx-theme#33 is merged

A preview:

versions

Depends on spyder-ide/spyder-docs-sphinx-theme#33

### Issue(s) Resolved

Part of #179


ccordoba12:
@dalthviz, please don't forget the way our docs are installed in Netlify to check how your work looks there.

CAM-Gerlach:
@dalthviz I'm not sure what specifically @ccordoba12 is referring to, but Netlify needs to be set to clone all branches like you did for Travis or only the current branch will show up in the dropdown on Netlify, as you can see in the preview. Not sure how best to do this with Netlify, but I'm sure you do.

ccordoba12:
> I'm not sure what specifically @ccordoba12 is referring to

To use `make multidocs` instead of `make docs` there.

> but Netlify needs to be set to clone all branches like you did for Travis or only the current branch will show up in the dropdown on Netlify

I don't know if that's possible because Netlify is in charge of cloning the repo (users don't control part).

dalthviz:
> To use make multidocs instead of make docs there.

For that we merged #186 and rebased this (since other PRs like #181 need to use `make docs` instead of `make multidocs` to generate a preview at least until we merge this PR).

> Netlify needs to be set to clone all branches

@CAM-Gerlach I was thinking about that too although, as @ccordoba12 says, I'm unsure if it's possible. I will need check

dalthviz:
Seems like Netlify builds a detached HEAD of the PRs without remotes:

image


Also thinking a little bit more, maybe having all the versions on the preview isn't actually needed right @CAM-Gerlach @ccordoba12 ? At the end when you are doing a modification you target that to only one version (depending on the base branch the PR sets) and I think Netlify puts your changes and builds that specific branch/version

ccordoba12:
> maybe having all the versions on the preview isn't actually needed right?

Sure, we can go without that for now.

ccordoba12:
I have some comments about the style though:

imagen

* I think we should style "Versions" in the same way as the rest of the text on the sidebar.
* "Versions" should be aligned with the title of the page or be a little below it.
* There's a margin on the left that prevents "Versions" and the dropdown to be aligned with "Search".
* We should style the dropdown in the same way as "Search".

----

My proposal (this also changes the border of "Search" a bit):

imagen

```css
#versions-header {
    color: rgba(0,0,0,.65);
    font-size: 0.8rem;
    text-transform: uppercase;
    margin-top: 1.1rem;
}
```

```css
#select-versions {
    margin-bottom: 4px;
    border: 1px solid rgba(0, 0, 0, 0.15)
    font-size: 90%;
    color: rgba(0,0,0,.65);
    padding: 2px;
}
```

```css
#select-versions, #select-versions option {
    width: 98%;
}
```

```css
.form-control {
    box-shadow: none;
    background: transparent;
    border: 1px solid rgba(0, 0, 0, 0.15);
    height: 54px;
    font-size: 18px;
    font-weight: 300;
}

**Comment by CAM-Gerlach** [comment]
It took way longer than it should, but I pushed my changes. I added a (overly fancy) Python script that generates the redirects and changed the `make multidocs` command accordingly, to first copy the current stable branch to a perma-URL (with a Python command, for maximum cross-platform compat), and then create the redirects to that. The selector still works, and selecting a version will take you to the URLs specific to that version, while simply using the stable URL or the base URL will go to the current stable branch so incoming links don't rot. or point to out of date versions, unless doing so is intentional.

In addition, I also fixed one major and one minor issue with the makefile (clean command not getting executed before `make multdocs`, causing the build to fail, and `multidocs` not being added to `.PHONY`), and refactored the CI config/scripts a bit to avoid needless code duplication between `ci/install.sh` and `ci/build.sh`, which isn't very DRY, increases maintenance burden, can result in them getting out of sync, and can increase the chance of errors/bugs (e.g. forgetting to update the theme repo/branch in one back to the stable before merging). Finally, I updated the Contributing Guide to reflect the various changes, including adding `sphinx-multiversion` to the install step, mentioning how to build the multiversion docs, and updating the Netlify config, plus some minor tweaks/reorganization.

Other than that, I think this should be ready to go, minus some minor style fixes which I can either make as suggestions or just commit directly, up to you @dalthviz , and updating Netlify to run both the install and build scripts.

File: doc/_static/custom_styles.css [hunk 1]
```diff
@@ -160,3 +160,44 @@ div#driver-popover-item .tour-progress-indicator {
...
+  width: 98%;
+}
+
+.form-control {
+  box-shadow: none;
+  background: transparent;
+  border: 1px solid rgba(0, 0, 0, 0.15);
+  height: 54px;
+  font-size: 18px;
+  font-weight: 300;
```
ccordoba12, line 202: @dalthviz, this change didn't take effect in Netlify. So, if every one agrees, I'm going to remove it and add it to the theme repo.

I was not expecting for my suggestions to be pushed by @dalthviz without discussion, but since it seems you liked them, I have no problem in doing that.

ccordoba12:
The warning bar on top is missing with these changes:

Selección_062

doc/_static/custom_styles.css
dalthviz: I did a change (load the custom css at the end) and now the changes are being applied, however it looks kind of strange. Could you check @ccordoba12 and maybe provide the correct style rules/missing style rules if needed? This is what I'm seeing using the Netlify preview:

image

If that is the correct proposal then I think it could look better if we drop the `font-size` and `height` rules and maybe add `text-transform` to uppercase. A preview:

With `text-transform`

image

Without `text-transform`

image

dalthviz:
The thing with the warning is that it will show in any page of any version. We need to check a better way to handle that and maybe an option is something we discussed with @CAM-Gerlach (which ended opening issue #188 ). Also, we decided with @CAM-Gerlach that the message was not needed since all this work is basically to release the documentation, right? (and if you need the older `3.x` docs you can use now the version select/dropdown)

CAM-Gerlach:
> At the end when you are doing a modification you target that to only one version (depending on the base branch the PR sets) and I think Netlify puts your changes and builds that specific branch/version

Yes, but fundamental to how Sphinx-Multiversion works is that with every PR, *all* the versions will be rebuilt with the build configuration (including the theme, conf.py, static HTML, CSS, JS and image assets) of the *current* version, so changes that touch any of that affect all of them (and have the potential to break things), so we need quickly way to easily verify that they all still work without major issues. Having Netlify build all the versions and allow us to easily check them with a few clicks would drastically reduce this otherwise considerable overhead.

doc/_static/custom_styles.css
ccordoba12: > I did a change (load the custom css at the end) and now the changes are being applied, however it looks kind of strange

@dalthviz, I think we shouldn't mess with that at this point, i.e. when we are about to do the final release. Not only our selected font is missing, but the style for paragraphs and links is different now:

Selección_065

I think we can explore this possibility after the release, but for now let's keep working as we've done it so far. Do you agree? And if so, could you revert the change you made?

CAM-Gerlach: Realized I never submitted this comment, sorry.

The problem is that `custom_styles.css` from the Docs repo (and the others listed in the custom styles key, as well as Pygments and Sphinx-Dropdown, which I'm _guessing_ is from Sphinx-Panels), gets injected _before_ the styles from the theme, `theme.css` and `spyder.css`, such that it doesn't override the theme's stylesheets as it should. They are injected [here](https://github.com/spyder-ide/spyder-docs-sphinx-theme/blob/6bec8d2434e783212c3347395fb6f8d8a19b2e30/pandas_sphinx_theme/layout.html#L18), while the theme.css and spyder.css are included directly below.

We could simply move the inject below the theme styles, [EDIT: That's what you did] but that wouldn't work on its own, because the `sphinx-bootstrap.css` stylesheet from `theme.conf` is also loaded in this stack and that would move it above the theme stylesheets where it shouldn't be; while our page doesn't actually use any of those styles (non-overriden at least, I've investigated every last one using Chrome's coverage and careful manual checks of both the page's appearance and the styles applied to every element it could potentially affect), it does use plenty from `basic.css` that this stylesheet `@import`s, and those need to be below the theme styles.

What I suggest instead that would preserve the current order in which styles are applied (except for the custom_styles CSS in question, which was always intended to go after the theme styles), as well as unify the theme variables (color, etc) as discussed last night, is `@import` or `@use` basic.css in `theme.css`, `@use` the `theme.scss` stylesheet in `spyder.scss`, remove the completely unused `sphinx-bootstrap.css` stylesheet and replace it with `spyder.css` in `theme.conf`, and remove the now-unneeded manual loading of `theme.css` and `spyder.css` in `layout.html`. Besides fixing the loading order of `custom_styles.css` without affecting the order of anything else important (to minimize risk), this allows use to use one set of variables (colors, font stacks) for all our theme stylesheets, as discussed previously, is more performant since it only requires one request instead of 4, keeps everything neat and seperate in the source repo.

@dalthviz , what do you think?

CAM-Gerlach:
> The thing with the warning is that it will be shown in any page of any version.

This is correct, I proposed a potential solution from the Sphinx-Multiversion docs but its complex enough to implement given our situation that we shouldn't block the release on it. Since we only have two branches and we have our redirects all set up, this won't be a huge issue, at least not right away.

> Also, we decided with @CAM-Gerlach that the message was not needed since all this work is basically to release the documentation, right?

This is the last step before the release, and merging this PR will, for all intents, _be_ the release since it will make our Spyder 4 docs fully publicly accessible and redirect to them by default.

doc/_static/custom_styles.css
dalthviz: I will revert the changes I did and update the `.form-control` style over the theme. About unifying the styles and rethinking the order we load them is something we need to do but after thinking about it maybe is out of the scope of this PR. Could you open an issue to refactor the style order and in general style handling, please @CAM-Gerlach ?

ccordoba12: I still don't understand why we `custom_styles.css` and `custom_scripts.js`. Why can't we unify everything in the Sphinx theme repo?

doc/_static/custom_styles.css
```diff
@@ -160,3 +160,35 @@ div#driver-popover-item .tour-progress-indicator {
...
+}
+
+#select-versions, #select-versions option {
+  width: 90%;
+}
+
+#select-versions {
+  margin: 10px;
+} */
+
```
ccordoba12, line 176:
```suggestion
```

Remove unused styles.

doc/_static/custom_styles.css
```diff
@@ -160,3 +160,35 @@ div#driver-popover-item .tour-progress-indicator {
...
+  text-transform: uppercase;
+  margin-top: 1.1rem;
+}
+
+#select-versions {
+  margin-bottom: 7px;
+  border: 1px solid rgba(0, 0, 0, 0.15);
+  font-size: 90%;
+  color: rgba(0,0,0,.65);
+  padding: 2px;
```
ccordoba12, line 189:
```suggestion
  padding: 2px;
  background-color: transparent;
```

An ugly gray background was shown in the dropdown.

### Before

Selección_066

### After

Selección_067

doc/_static/custom_styles.css
dalthviz: `custom_styles.css` manages styles that affect specific elements of the content that is being added in Spyder docs like the color of the icons/symbols the videos added with the directive and the interactive tutorial I think. `custom_scripts.js` was created to handle the tour load (and there I added too the handling of the version selector/dropdown).

ccordoba12: Ok, that makes a lot of sense. However, after looking at `custom_styles.css` I think some style declarations should be moved to the Sphinx repo instead, such as `iframe`, `#versions-header`, `#select-versions` and `#select-versions, #select-versions option`.

What do you think?

dalthviz: Maybe `iframe` could be moved but the elements referring to versions are over this side (the template that renders the versions selector/dropdown is here)

dalthviz: Should I move the `iframe` rules to the theme @ccordoba12 ?

CAM-Gerlach: > maybe is out of the scope of this PR. Could you open an issue to refactor the style order and in general style handling, please @CAM-Gerlach ?

I agree, opened spyder-ide/spyder-docs-sphinx-theme#34

> Why can't we unify everything in the Sphinx theme repo?

For the same reason the theme is separate from the spyder-docs repo in the first place, because the purpose of a theme is to be modular and consumable by multiple projects (e.g. our future API docs, our other projects' docs, external projects, etc.) and to keep the core theming separate from the bespoke styles specific to the particular content we have at any given time.

> I think some style declarations should be moved to the Sphinx repo instead, such as iframe, #versions-header, #select-versions and #select-versions, #select-versions option.

At least to my recollection, the `iframe` rules are fairly specific to our use case and to our custom directive and container divs that implements them, that are hosted here in the theme.

As for the last three, it doesn't make sense to move the CSS styles without moving the version selector script and the HTML template, since they explicitly depend on them to do anything useful. I originally suggested this to @dalthviz on the basis that the version selector could be a useful common resource for other theme consumers, like the API docs, rather than something bespoke to our use case, but @dalthviz brought up that it would be "dead code" without the `sphinx-multiversion` dependency that not all theme users might require, and that if and when we need it elsewhere in the future, we could move it to the theme later.

Personally, I don't have a problem with support for optional dependencies in the theme, but since things are working as they are now and this is the final step to make the release, I'm not sure its worth blocking over a debate that doesn't affect the user-visible output and that we can easily resolve later.

CAM-Gerlach:
> Until we decide how to handle the "Versions" dropdown in that situation, I think we should only show the sidebar one.

The problem is this leaves users on smaller screens/mobile without a way to change the version at all...can't we just add it as another dropdown above or below the sidebar one?

dalthviz:
@ccordoba12 @CAM-Gerlach could you check again? Seems like I was missing a pull for the branches to actually show the most updated version of each branch (you can choose `master`, 3.x or `pr`). The way the selectors/dropdowns are currently working is that in small screens both are being shown

doc/_static/custom_styles.css
ccordoba12: Ok, fair enough. Let's leave the css rules I mentioned here for the time being.

It's not a big issue and certainly not a blocker before the release.

ccordoba12:
> @ccordoba12 @CAM-Gerlach could you check again?

Things are working fine now. Thanks @dalthviz!

ccordoba12:
I have another suggestion. The dropdown is showing these entries in the latest commit pushed here (https://github.com/spyder-ide/spyder-docs/pull/185/commits/4701420ab5c9c49faf93ac445bf39118d6536a24):

Selección_072

Could it be possible to rename them to `Spyder 3` (instead of `3.x`) and `Spyder 4` (instead of `current`)? If that's too hard to do before the release, then there's no problem. But I think it's kind of hard for users to associate those names with specific content.

CAM-Gerlach:
@ccordoba12 As I've mentioned previously (somewhere... :) ) I've been working on that for a while and will push my changes shortly. It ended up being way more complicated to both do that and also build the current feature/PR branch, because I was having to set the release in conf.py dynamically (in order to also rename the output directories accordingly) and couldn't figure out any way to propagate the right state or env variables all the way through to the sub-build to get it to work reliably cross-platform but I think I've got it solved now.

Makefile
```diff
@@ -33,15 +33,22 @@ autodocs: ## generate Sphinx HTML documentation, including API docs
...
+	@python scripts/generateredirects.py "current" -v --base-path "doc/_build/html" --base-url "https://docs.spyder-ide.org"
+
 linkcheck: ## check that links are still valid
 	$(MAKE) -C doc linkcheck
 
 servedocs: doc ## compile the docs watching for changes
 	watchmedo shell-command -p '*.rst' -c '$(MAKE) -C doc html' -R -D .
 
-serve: clean ## install the package to the active Python's site-packages
+serve: clean ## Launch the docs on the browser
```
CAM-Gerlach, line 53:
```suggestion
serve: clean ## Launch the docs in a web browser
```

doc/_static/custom_scripts.js
```diff
@@ -175,6 +175,42 @@
...
+    
+    // Handle version selector
+    function setupVersionSelector() {
+        var selectOptions = document.querySelectorAll("#select-versions option");
+        var currentPage = window.location.pathname.split("/").slice(-1)[0];
+        var selection = [...selectOptions].filter(function(elem){
+            return elem.value === currentPage;
+        });
+
+        if(selection.length) {
```
CAM-Gerlach, line 187:
```suggestion
        if (selection.length) {
```

doc/conf.py
```diff
@@ -43,6 +44,7 @@
 extensions = [
     "sphinx.ext.githubpages",
     "sphinx_panels",
+    "sphinx_multiversion"
```
CAM-Gerlach, line 47:
```suggestion
    "sphinx_multiversion",
```

doc/_static/custom_scripts.js
```diff
@@ -175,6 +175,42 @@
...
+            });
+        }
+
+        // Based on https://stackoverflow.com/a/37796085
+        var isSafari = navigator.userAgent.indexOf("Safari") != -1;
+        var isChrome = navigator.userAgent.indexOf("Chrome") != -1;
+        if (isChrome || isSafari){
+          document.querySelectorAll("#select-versions").forEach(function(ele) {
+            ele.onchange = function () {
+                if(this.value) {
```
CAM-Gerlach, line 199:
```suggestion
                if (this.value) {
```

doc/_static/custom_scripts.js
```diff
@@ -175,6 +175,42 @@
...
+                  window.location.href = this.value;
+                }
+            };
+          });
+        } else {
+          document.querySelectorAll("#select-versions option").forEach(function(ele) {
+            ele.onclick = function () {
+                if(this.value) {
```
CAM-Gerlach, line 207:
```suggestion
                if (this.value) {
```

doc/_static/custom_scripts.js
```diff
@@ -175,6 +175,42 @@
...
+        if(selection.length) {
+            selection.forEach(function(elem){
+                elem.setAttribute("selected", true);
+            });
+        }
+
+        // Based on https://stackoverflow.com/a/37796085
+        var isSafari = navigator.userAgent.indexOf("Safari") != -1;
+        var isChrome = navigator.userAgent.indexOf("Chrome") != -1;
+        if (isChrome || isSafari){
```
CAM-Gerlach, line 196: As I'm sure you're aware, manually probing the user agent of the browser like this and changing behavior accordingly is a strongly frowned upon practice these days, in favor of checking for support of the specific features directly. It might not work on less common browsers, or if the user agent string is changed, or in many other scenarios, and seems quite fragile. Are you sure there isn't a better way of doing this?

ci/install.sh
```diff
@@ -1,5 +1,14 @@
 #!/bin/bash -ex
 
 pip3 install -U pip
-pip3 install sphinx doctr sphinx-panels
-pip3 install git+https://github.com/spyder-ide/spyder-docs-sphinx-theme.git@develop_spyder
+pip3 install sphinx doctr sphinx-panels sphinx-multiversion
+# You can change the command above to use a different version of the theme (but return it to the correct one before merging)
+pip3 install git+https://github.com/dalthviz/spyder-docs-sphinx-theme.git@sidebar_widgets
```
CAM-Gerlach, line 6: REMINDER: CHANGE BEFORE MERGING!!!

doc/_static/custom_scripts.js
```diff
@@ -185,11 +221,13 @@
         };
     });
 
-    // Once everything is loaded, start the tour
+    // Once everything is loaded, start the tour, handle version dropdown
```
CAM-Gerlach, line 224:
```suggestion
    // Once everything is loaded, start the tour and handle the version dropdown
```

doc/_static/custom_scripts.js
CAM-Gerlach: @dalthviz

CAM-Gerlach:
@dalthviz , @juanis2112 noticed that in Safari, the selector functions, but displays the wrong branch (always the current one) once the page is loaded. In Chrome, we can both confirm that it starts off on the default selection on page load, but within a couple hundred ms is set to the correct one, and in FF it is always correct from when the page first starts to render, but in Safari it says set to the current branch permanently.

image

doc/_static/custom_scripts.js
dalthviz: Will check

doc/_static/custom_scripts.js
```diff
@@ -186,30 +186,17 @@
...
-            ele.onclick = function () {
-                if (this.value) {
-                  window.location.href = this.value;
-                }
-            };
-          });
-        }
+        document.querySelectorAll("#select-versions").forEach(function(ele) {
+          ele.onchange = function () {
+              if(this.value) {
```
CAM-Gerlach, line 195:
```suggestion
              if (this.value) {
```

doc/_static/custom_scripts.js
CAM-Gerlach: Looks like this was resolved, thanks @dalthviz

CAM-Gerlach:
@juanis2112 I fixed the "blink" issue with the version selector, and made a few related behind the scenes improvements

@dalthviz A few last changes:

* Make the `setupVersionSelector()` function fire when the DOM is ready, instead of only when all external resources are loaded, to ensure it triggers earlier to reduce the window in which it is non-functional for users on slow connections (and originally to reduce or eliminate the blink, but I fixed it entirely with the next change
* Statically set the `option` that is `selected` via a check in the Jinja template, and eliminate the now-unnecessary JS that sets it, so that it is set correctly from the very beginning and avoid the need for the extra complexity, fragility and delay it introduces
 * Only run `setupVersionSelector()` if there is a version selector on the page (e.g. `make docs`; I didn't observe any console errors or bad behavior but better safe than sorry)
* Minor style tweaks and kill blank spaces

CAM-Gerlach:
@juanis2112 rechecked and it looks good, so @dalthviz you're free to go ahead with the above, thanks.

PR #191, opened 2020-08-30
PR: Set doctr deploy directory to root per multiversion changes

CAM-Gerlach:
## Description of Changes

Doctr was previously deploying `master` to the `develop` subdir of the site, but with the multiversion changes, it needs to deploy to the root.


CAM-Gerlach:
I'm going to go ahead and merge this since @dalthviz agrees with the fix, the builds pass and there's no other way to test this.

PR #205, opened 2020-11-29
PR: Add FAQ questions on Spyder being open source and Anaconda situation

### LINKED ISSUE #202: Add FAQ question on Spyder's license and the impact of the Anaconda EULA change
## Issue Description

We've gotten a number of questions lately (and we've always had a few) about whether Spyder is free for commercial use and distribution, and particularly what the impact of the [Anaconda EULA change](https://www.anaconda.com/blog/sustaining-our-stewardship-of-the-open-source-data-science-community) is on it, now that it seems to have come to the fore. Therefore, it would be good to have a short FAQ question explaining this aspect, based on some of the responses in our Gitter.

Source material to draw from:

> To provide a little more detail, Spyder is community developed and 100% free and open source (there is no paid version or Spyder company, like with most other IDEs, though we do have generous sponsors that help fund our development and specific features). Our source code, standalone installers and most of our distribution methods (pip/PyPI, Linux distros, MacPorts, WinPython, etc) are fully unemcumbered and can be freely redistributed, used and modified by anyone.
> 
> Full-fat Anaconda proper does come with restrictions for big for-profit companies using it on a large scale, but that restriction is only on the package distribution infrastructure (the packaged Anaconda distribution and the `defaults`conda channel); instead, you can simply download the more lightweight [Miniconda](https://docs.conda.io/en/latest/miniconda.html) which is essentially just Anaconda without all the gigabytes of Python package packages installed by default and is 100% open source and free of restrictions, and install whatever you need (including Spyder and any packages you want to work with) from the community maintained `conda-forge` channel, which has a much wider variety of packages and is generally more up to date than the Anaconda `defaults` channel. I actually prefer this approach for the mentioned reasons  and is what I've used for years, despite working as a NASA-funded researcher in academia.

> To clarify this, Spyder now offers standalone installers for Windows and macOS, which don't rely on Anaconda and are 100% free and open source; the download button on our website now points to them instead on the appropriate platforms. And in terms of conda, the situation is actually a good deal more nuanced than this; Anaconda has made statements to the effect that it only applies if you're using it _at scale_ in your large company; if only a few people are actually using it or on a part-time basis, they're not going to go after you. Furthermore, conda itself, Miniconda and Conda-Forge are still 100% free and open source, so its perfectly possible to still use it just as you currently do, with only a few changes to your workflow.

> They can if they install Spyder in its own environment rather than `base`, as has been our official recommendation for years now, and use conda-forge for that env instead of `defaults` (not sure if CF Spyder and Spyder-Kernels will be have compat issue with an all-AD stack beneath it; Spyder itself is pure Python so I'd _think_ not but some packages might  get channel-switched piecemeal so an all-CF Spyder env would naturally be the officially supported solution). Since users don't need to reinstall any of their packages, just Spyder and any optional deps they want to use with it (Numpy, Pandas, Matplotlib, etc), its relatively cheap to simply create a new env for CF Spyder if they happen to haveo one already.

FAQ

### Other information

Yes, Spyder 4.x

Yes

CAM-Gerlach:
### Pull Request Checklist

## Description of Changes

Adds a new section, "About Spyder", with two new FAQ questions: one clarifying that Spyder is open source and fully free to use, one explaining how to use it without running in to the Anaconda EULA restrictions on large commercial companies.

image

### Issue(s) Resolved

Fixes #202

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -45,28 +45,6 @@ Installing and updating
       :alt: Navigator showing installing specific version of Spyder
 
 
-.. dropdown:: Q: Is commercial use allowed? What's the situation with Anaconda?
-   :container: + dropdown-id-commercial-use
-
-   Spyder is 100% free and open source; there is no paid version or prohibition on commercial use.
-   It is developed by its international user community, and supported by its users through `OpenCollective`_ and by its generous sponsoring organizations, including `Quansight`_ and `NumFOCUS`_.
-   Our source code, standalone installers and most of our distribution methods (Pip/PyPI, Linux distros, MacPorts, WinPython, etc) can be freely redistributed, used and modified by anyone, for any purpose, including commercial use.
-
-   .. _OpenCollective: https://opencollective.com/spyder
-   .. _Quansight: https://www.quansight.com/
-   .. _NumFOCUS: https://numfocus.org/
-
-   The Anaconda distribution `recently changed`_ their `Terms of Service`_ to add restrictions on large (> 200 employee) for-profit companies using it on a large scale, but that only applies to the package distribution infrastructure (the packaged Anaconda distribution and the ``defaults`` conda channel).
-   Instead, you can simply download the similar `Miniforge distribution`_, which is 100% open source and identical to full Anaconda except for the many Python packages installed by default in the Anaconda ``base`` environment (which we recommend you avoid using anyway, as any problems here can break your whole Anaconda installation).
-   Then, simply install the packages you need (including Spyder, if you aren't using our recommended :ref:`standalone_installers_ref`) with ``conda`` as you usually do.
-   Miniforge will automatically use the community-maintained Conda-Forge repository, which has a much wider variety of packages and is generally more up to date than the Anaconda equivalent, in addition to being free of any commercial restrictions.
-   For more, see our :doc:`installation`.
-
-   .. _recently changed: https://www.anaconda.com/blog/sustaining-our-stewardship-of-the-open-source-data-science-community
-   .. _Terms of Service: https://www.anaconda.com/terms-of-service
-   .. _Miniforge distribution: https://github.com/conda-forge/miniforge/releases/latest
-
-
 
 ==============
 Running Spyder
@@ -343,6 +321,40 @@ Troubleshooting
 
    .. code-block:: bash
 
-       conda activate spyder-dev
-       export QT_MAC_WANTS_LAYER=1
-       spyder
+      conda activate spyder-dev
+      export QT_MAC_WANTS_LAYER=1
+      spyder
+
+
+
+============
+About Spyder
+============
+
+.. dropdown:: Q: What's Spyder's licensing situation? Is commercial use allowed?
+   :container: + dropdown-id-commercial-use
+
+   Spyder is 100% free and open source; there is no paid version or prohibition on commercial use.
+   It is developed by its international user community, and supported by its users through `OpenCollective`_ and by its generous sponsoring organizations, including `Quansight`_ and `NumFOCUS`_.
+   Our source code, standalone installers and most of our distribution methods (Pip/PyPI, Linux distros, MacPorts, WinPython, etc) can be freely redistributed, used and modified by anyone, for any purpose, including commercial use.
+   For more details about the situation with Anaconda, see `that question`_.
+
+   .. _OpenCollective: https://opencollective.com/spyder
+   .. _Quansight: https://www.quansight.com/
+   .. _NumFOCUS: https://numfocus.org/
+   .. _that question: #anaconda-license
+
+
+.. dropdown:: Q: What do the Anaconda licensing changes mean for Spyder?
+   :container: + dropdown-id-anaconda-license
+
+   If you use Spyder with the Anaconda distribution, they `recently changed`_ their `Terms of Service`_ to add restrictions on large (> 200 employee) for-profit companies using Anaconda on a large scale.
+   However, these terms only apply to the package infrastructure (the full Anaconda distribution and the ``defaults`` conda channel).
+   Instead, you can simply download the similar `Miniforge distribution`_, which is 100% open source and identical to full Anaconda (aside from not bundling the Python packages installed by default in the Anaconda ``base`` environment, which we recommend you avoid using anyway given any problems here can break your whole installation).
+   Then, simply install the packages you need (including Spyder, if you aren't using our recommended :ref:`standalone_installers_ref`) with ``conda`` as you usually do.
+   Miniforge will automatically use the community-maintained Conda-Forge repository, which has a much wider variety of packages and is generally more up to date than the Anaconda equivalent, in addition to being free of any commercial restrictions.
+   For more, see our :doc:`installation`.
+
+   .. _recently changed: https://www.anaconda.com/blog/sustaining-our-stewardship-of-the-open-source-data-science-community
+   .. _Terms of Service: https://www.anaconda.com/terms-of-service
+   .. _Miniforge distribution: https://github.com/conda-forge/miniforge/releases/latest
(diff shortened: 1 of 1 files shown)


doc/faq.rst
```diff
@@ -42,6 +42,25 @@ Installing and updating
       :alt: Navigator showing installing specific version of Spyder
 
 
+.. dropdown:: Q: Is commercial use allowed? What's the situation with Anaconda?
+
+   Spyder is 100% free and open source; there is no paid version or prohibition on commercial use.
+   Spyder is developed by its international user community, and supported by its users through `OpenCollective`_ and by its generous sponsoring organizations, including `Quansight`_ and `NumFOCUS`_.
+   Our source code, standalone installers and most of our distribution methods (Pip/PyPI, Linux distros, MacPorts, WinPython, etc) are fully unencumbered and can be freely redistributed, used and modified by anyone.
```
juanis2112, line 49: I would suggest replacing the "unencumbered" word for something more familiar (specially for non native speakers).

juanis2112, line 49: I would suggest replacing the "unencumbered" word for something more familiar (specially for non native speakers).

doc/faq.rst
```diff
@@ -42,6 +42,25 @@ Installing and updating
...
+
+   .. _OpenCollective: https://opencollective.com/spyder
+   .. _Quansight: https://www.quansight.com/
+   .. _NumFOCUS: https://numfocus.org/
+
+   The Anaconda distribution recently `recently changed`_ their EULA to add restrictions on large (> 200 employee) for-profit companies using it on a large scale, but that only applies to the package distribution infrastructure (the packaged Anaconda distribution and the ``defaults`` conda channel).
```
juanis2112, line 55:
```suggestion
   The Anaconda distribution recently `changed`_ their EULA to add restrictions on large (> 200 employee) for-profit companies using it on a large scale, but that only applies to the package distribution infrastructure (the packaged Anaconda distribution and the ``defaults`` conda channel).
```
Repeated word

juanis2112 (changes requested):
I think it might be better to separate these two in two different questions in the FAQ as they are not actually too related and it makes the answer too long to have them like this. Also if we do this, we can create another section that can be called like "About Spyder" or something like that, since from the question the connection to the section is not very clear. 

I would also suggest, if we split these questions up, making them more specific and clear cause for instance the "What's the situation with Anaconda" one it's not very clear as in... what situation specifically you are talking about.

doc/faq.rst
```diff
@@ -42,6 +42,25 @@ Installing and updating
...
+   Instead, you can simply download the similar `Miniconda distribution`_, which is 100% open source and identical to full Anaconda except for the gigabytes of Python packages installed by default in the Anaconda ``base`` environment (which we recommend you avoid using anyway).
```
ccordoba12, line 56:
```suggestion
   Instead, you can simply download the similar `Miniconda distribution`_, which is 100% open source and identical to full Anaconda except for the many Python packages installed by default in the Anaconda ``base`` environment (which we recommend you avoid using anyway).
```

- The phrasing seems derogatory, which reads opinionated and not something I think we should include in our FAQ.
- Also, perhaps we should point directly to Miniforge instead of Miniconda?

doc/faq.rst
```diff
@@ -42,6 +42,25 @@ Installing and updating
...
+   Then, simply install whatever packages you need (including Spyder, if you aren't using our recommended stand-alone installers) from the community-maintained ``conda-forge`` channel, which has a much wider variety of packages and is generally more up to date than the Anaconda ``defaults`` channel.
```
ccordoba12, line 57:
```suggestion
   Then, simply install the packages you need (including Spyder, if you aren't using our recommended stand-alone installers) from the community-maintained ``conda-forge`` channel, which has a much wider variety of packages and is generally more up to date than the Anaconda ``defaults`` channel.
```

- I don't like `whatever` too much here.
- I think we should also add a link to the section in our Installation guide about our new installers.

ccordoba12 (commented):
Besides my suggestions below, I also agree with @juanis2112's very good suggestions: we should split the text in two questions and add a new secton for them.

My suggestion for the header of that section: `Spyder and Anaconda`

doc/faq.rst
CAM-Gerlach:
```suggestion
   Our source code, standalone installers and most of our distribution methods (Pip/PyPI, Linux distros, MacPorts, WinPython, etc) can be freely redistributed, used and modified by anyone, for any purpose, including commercial use.
```

Yeah, I don't know why I thought that was a good idea in a public FAQ for beginners from around the world. Ended up removing it completely and instead describing what I meant more clearly and explicitly.

doc/faq.rst
CAM-Gerlach:
```suggestion
   The Anaconda distribution `recently changed`_ their EULA to add restrictions on large (> 200 employee) for-profit companies using it on a large scale, but that only applies to the package distribution infrastructure (the packaged Anaconda distribution and the ``defaults`` conda channel).
```

Avoid breaking the link reference

doc/faq.rst
CAM-Gerlach: > The phrasing seems derogatory, which reads opinionated and not something I think we should include in our FAQ.

Agreed

> Also, perhaps we should point directly to Miniforge instead of Miniconda?

Absolutely, didn't even know that existed. I'll do it when I pull and rebase these in, since I'll need to change the link target as well

doc/faq.rst
CAM-Gerlach: > I don't like whatever too much here.

Agreed

> I think we should also add a link to the section in our Installation guide about our new installers.

Good point, I'll do that in the rebase as well

doc/faq.rst
CAM-Gerlach: Done; also modified the follow sentences to account for that change.

doc/faq.rst
CAM-Gerlach: Added

CAM-Gerlach:
@ccordoba12 @juanis2112 Rebased in the links, added the class ids, applied your changes, made the additional changes @ccordoba12 suggested, split the questions and added them to a new section as @juanis2112 , and fixed/improved the wording and formatting in a number of places.

ccordoba12:
@CAM-Gerlach, this has a merge conflict now.

CAM-Gerlach:
@ccordoba12 Rebased to fix.

CAM-Gerlach:
@ccordoba12 @juanis2112 Bump

juanis2112:
Will review now, sorry

doc/faq.rst
```diff
@@ -321,6 +321,40 @@ Troubleshooting
...
+   .. _Quansight: https://www.quansight.com/
+   .. _NumFOCUS: https://numfocus.org/
+   .. _that question: #anaconda-license
+
+
+.. dropdown:: Q: What do the Anaconda licensing changes mean for Spyder?
+   :container: + dropdown-id-anaconda-license
+
+   If you use Spyder with the Anaconda distribution, they `recently changed`_ their `Terms of Service`_ to add restrictions on large (> 200 employee) for-profit companies using Anaconda on a large scale.
+   However, these terms only applies to the package infrastructure (the full Anaconda distribution and the ``defaults`` conda channel).
```
juanis2112, line 352:
```suggestion
   However, these terms only apply to the package infrastructure (the full Anaconda distribution and the ``defaults`` conda channel).
```
Isn't it like this?

juanis2112 (approved):
I think this looks good now

doc/faq.rst
CAM-Gerlach: Yes indeed, good catch. I'll fixup it in.

CAM-Gerlach: Done

ccordoba12 (approved):
It looks pretty good @CAM-Gerlach! Thanks for your work on this.

PR #206, opened 2020-11-29
PR: Add FAQ question on installing Spyder on macOS Big Sur

CAM-Gerlach:
### Pull Request Checklist

* [x] Read and followed this repo's [Contributing Guidelines](https://github.com/spyder-ide/spyder/blob/master/CONTRIBUTING.md)
* [x] Based your PR on the latest version of the correct branch (master or 3.x)
* [x] Checked your writing carefully for correct English spelling, grammar, etc
* [x] Described your changes and the motivation for them below
* [x] Noted what issue(s) this pull request resolves, creating one if needed

## Description of Changes

Adds a FAQ question describing the macOS Big Sur situation and explaining how to work around it, to be linked to from the website download section (once a PR that implements #204 is implemented).

image

### Issue(s) Resolved

Fixes #203

Issue #203, opened 2020-11-29
Add FAQ question on troubleshooting Big Sur incompatibility to link to from download section of site


CAM-Gerlach:
## Issue Description

There's been a lot of people on various channels trying to install Spyder on macOS Big Sur, where it is not yet supported and won't work yet, unless users follow the workaround proposed by @ccordoba12 in [this comment](https://github.com/spyder-ide/spyder/issues/14218#issuecomment-730141294) on spyder-ide/spyder#14218 and [this comment](https://github.com/spyder-ide/spyder/issues/14222#issuecomment-730139263) on spyder-ide/spyder#14222 . Until our installers work with it, we should have an FAQ question that explains this, and include a prominent link to it in our downloads section of the site (I could link the button to the FAQ directly if it detects Big Sur with a bit of Javascript, but that could be finicky and is a non-trivial amount of work for a change that will only be up a few weeks.

Needs #204 to actually be able to link to the question, and depending on how I do the `id`s, I might need to to rebase the other PR after whichever one of these goes in first.

### Where is the relevant portion of the documentation located?

FAQ, Troubleshooting section

### What would you suggest it be replaced with?

Raw material from @ccordoba12 's aforementioned comments:

> Sorry guys but I'm going to lock this issue because it's starting to accumulate incorrect answers (some of which are very bad advice).
> 
> Bottom line is this:
> 
> * Spyder will work with Big Sur in our next version (4.2.1), to be released in one month.
> 
> * The workaround for now is described in the last comment on issue #14218.
> 
> * Our Mac app is going to be updated in 4.2.1 too to work with Big Sur. Please don't mess with it because it will defeat its purpose of providing a very stable way to deliver Spyder to users.
> 
> * Additional issues about Big Sur will be closed and people will be pointed to issue #14218 for the workaround.

> If you prefer to use only conda packages, the solution consists first in installing Spyder with:
> 
> ```
> conda create -n spyder-dev python=3
> conda activate spyder-dev
> conda install -c conda-forge spyder
> ```
> 
> Then, to start Spyder and avoid latency issues, you need to run
> 
> ```
> conda activate spyder-dev
> export QT_MAC_WANTS_LAYER=1
> spyder
> ```

### Other information

* Is the problem specific to documentation for a particular version of Spyder?

Yes, 4.2.0 only

* Are you planning on submitting a Github pull request with a suitable change?

Yes

PR #207, opened 2020-11-30
PR: Add anchors, direct links and buttons to dropdowns, and open and scroll to them automatically

### LINKED ISSUE #204: Add support for direct links to individual FAQ questions that expand them
## Issue Description

We need to be able to directly link to the FAQ question to be added per #203 from the website, and this would be very useful to have to link users elsewhere to FAQ questions, especially as the page continues to grow. However, FAQ questions don't have unique ids that I can link to directly, nor can I get them to open automatically, so the nearest thing that can be currently linked to is the section. Therefore, this is a priority to add.

FAQ page

After some investigation and testing I can add some JS to do that, since that would be extremely helpful in a lot of other contexts to where we want to link to FAQ questions. I also opened an issue, executablebooks/sphinx-panels#56 on Sphinx-Panels to add an `:id:` property that will allow us to more easily do that without crude hacks (for now I can work around it with one-off classes and some more hacky JS). I will also add a FA icon from where you can easily copy said link.

We could just scroll to the element by ID, but that wouldn't take into account the navbar and leave the question obscured at the top of the page, and also wouldn't actually open the linked question. Instead, we can pass the question id as a param and use a bit more JS to scroll to it minus some vertical offset to account for the navbar and give it a little more room, and automatically open the dropdown section, which should be streightforward enough to do after some testing.

### Other information

No

Yes

* Anything else you'd like to add?

Example of link button, as other FAQs and similar have:

image

CAM-Gerlach:
### Pull Request Checklist

## Description of Changes

Adds id anchors to each dropdown (e.g. FAQ question), allowing them to be linked to directly, along with JS that automatically scrolls to them, adjusts for the navbar and gives them some margin, and opens them. I also added link widgets to each one that behave and are styled similarly to the ones next to section headers, that allow easily copying and pasting the links, and bespoke custom classes to each FAQ question to give each a short, unique and permanently linkable name. This is specifically needed for issue #203 and PR #206 so that the FAQ questions can be linked for the side, but should be very useful in general so we can easily link users asking common questions directly to them, instead of just to the section.

Once executablebooks/sphinx-panels#56 is merged, this can be simplified to remove much of the custom setup code that relies on bespoke classes and JS hacks, and use names/ids directly instead, but I've designed the code and docs so the transition, when it happens, won't actually affect the links and will only need relatively straightforward changes.

Either this will need to be merged first and PRs #205 and #206 rebased to add the class/id names, or vice versa.

### Issue(s) Resolved

Fixes #204

### DIFF
--- a/doc/_static/custom_scripts.js
+++ b/doc/_static/custom_scripts.js
@@ -163,12 +163,12 @@
     function getDropdownElement() {
         var dropdownID = window.location.hash;
         if (! dropdownID) {
-          return false;
+            return false;
         };
 
         var dropdownElement = document.getElementById(dropdownID.substring(1));
         if ((! dropdownElement) || (! dropdownElement.classList.contains(dropdownClassName))) {
-          return false;
+            return false;
         };
 
         return dropdownElement;
@@ -210,7 +210,14 @@
     function setupDropdownLinks() {
         var dropdowns = document.getElementsByClassName(dropdownClassName);
         for (var i = 0; i < dropdowns.length; i++) {
-            dropdowns[i].id = "dropdown-" + (i + 1);
+            for (var j = 0; j < dropdowns[i].classList.length; j++) {
+                if (dropdowns[i].classList[j].startsWith("dropdown-id-")) {
+                    dropdowns[i].id = dropdowns[i].classList[j].replace("dropdown-id-", "");
+                };
+            };
+            if (! dropdowns[i].id) {
+                dropdowns[i].id = "dropdown-" + (i + 1);
+            };
 
             var aTag = document.createElement('a');
             aTag.setAttribute("href", "#" + dropdowns[i].id)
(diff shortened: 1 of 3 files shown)


ccordoba12 (approved):
Tried on Netlify and it's working great, thanks @CAM-Gerlach!

CAM-Gerlach:
Added a minor tweak that should improve reliability of the scrolling. Otherwise, this can be merged once @juanis2112 tests it and is happy with the style, since I can just add the right directive to the two FAQ PRs so they'll have their unique anchor ids once they get merged.

Style-wise, as a baseline I made the link anchors match the section header anchor links that serve the same function, though I'm not sure I'm sold on the look. If you all prefer, we could use our blue color instead, which would look less out of place with the FAQ its part of (at the cost of being less cohesive with the other links).

juanis2112 (approved):
I'll suggest using the blue color for the links

CAM-Gerlach:
Done, it looks _way_ better in with the blue. Unless either of you have any last-minute objections, I'll go ahead and merge once the checks pass and I verify Netlify looks good, since we agreed on it otherwise and both of you approved it.

PR #258, opened 2021-05-28
PR: Update openSUSE link

bnavigator:
spyder 3 is the old name, the new name of the RPM package is spyder.


ccordoba12 (approved):
Thanks @bnavigator!

PR #281, opened 2021-09-02
Added spyder-kernels version for spyder 5.1.*

ba-tno:
# Pull Request

## Pull Request Checklist

## Description of Changes
The documentation was missing a line for the `spyder-kernels` version of the current 5.1.* Spyder version, I've added that in the current pull request.

## Issue(s) Resolved

Fixes #

### DIFF
--- a/doc/troubleshooting/common-illnesses.rst
+++ b/doc/troubleshooting/common-illnesses.rst
@@ -40,6 +40,7 @@ Check the required version of spyder-kernels for your version of Spyder in the f
    4.1.5-4.1.6      1.9.4
    4.2.0            1.10.0
    5.0.0-5.0.5      2.0.1-2.0.5
+   5.1.*            2.1.*
    ==============   ==============
 
 To do so, activate the environment, then install ``spyder-kernels``.


juanis2112 (approved):
Thanks @ba-tno for adding this.

juanis2112:
@ccordoba12 I think we can merge this one. The test not passing is from a previous PR.

doc/troubleshooting/common-illnesses.rst
```diff
@@ -40,6 +40,7 @@ Check the required version of spyder-kernels for your version of Spyder in the f
    4.1.5-4.1.6      1.9.4
    4.2.0            1.10.0
    5.0.0-5.0.5      2.0.1-2.0.5
+   5.1.*            2.1.*
```
CAM-Gerlach, line 43:
```suggestion
   5.0.0-5.0.5      2.0.5
   5.1.0-5.1.5      2.1.1
```

See review comment for details; thanks.

CAM-Gerlach (changes requested):
Thanks for the PR, @ba-tno , good catch.

Unfortunately, as Spyder's versioning scheme enforces relatively tight coupling between Spyder and Spyder-Kernels versions, Spyder patch versions are not guaranteed to not affect compatibility with Spyder-Kernels patch versions; this is usually to ensure critical bugs are fixed. This was the case with Spyder 5.1.2, which requires Spyder-Kernels 2.1.1+. Likewise, future A.B.x Spyder versions cannot be assumed to be compatible with any C.D.x Spyder-Kernels version, so unfortunately while it does require more maintenance, the version ranges need to be specified explicit.

Also, the motivating use of the Spyder-Kernels version is as input to the following `conda install`/`pip install` command, so its effectively pinned to that version, and thus if a specific patch version is required, we have to explicitly list the latest version (even though, again, that's not really ideal; the approach you suggest would be better if patch version compatibility was maintained).

doc/troubleshooting/common-illnesses.rst
ccordoba12: I agree with this suggestion.

CAM-Gerlach (approved):
LGTM, thanks @ba-tno ! Since the change was :+1: ed by @ccordoba12 and previously :heavy_check_mark: ed by @juanis2112 , and I confirmed everything displays correctly in the preview and none of the check failures were introduced by this PR (I'll fix those shortly) I'll go ahead to squash and merge.

CAM-Gerlach:
I just noticed that this table is apparently [duplicated](https://docs.spyder-ide.org/current/faq.html#troubleshooting-starting-kernel) in the FAQ question that links here. To avoid having to constantly update and maintain both tables and risk one or both drifting out of sync/out of date, we should just link this table directly from that question (along with the question itself already linking to this section). I can open a followup PR.

ccordoba12:
Agreed.

PR #277, opened 2021-08-02
PR: Add using additional modules video to FAQ in docs

juanis2112:
# Pull Request

## Pull Request Checklist

* [ ] Read and followed this repo's [Contributing Guidelines](https://github.com/spyder-ide/spyder/blob/master/CONTRIBUTING.md)
* [ ] Based your PR on the latest version of the correct branch (master or 4.x)
* [ ] Checked your writing carefully for correct English spelling, grammar, etc
* [ ] Described your changes and the motivation for them below
* [ ] Noted what issue(s) this pull request resolves, creating one if needed

## Description of Changes

Added video on  using additional module to FAQ in docs.

## Issue(s) Resolved

Fixes #269

PR #282, opened 2021-09-17
PR: Fix standalone installers and packages dropdown in FAQ showing id as text

dalthviz:
# Pull Request

## Pull Request Checklist

## Description of Changes

The id of the dropdown is being shown as part of the text of the item:

image

## Issue(s) Resolved

Fixes #

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -176,7 +176,7 @@ Using Spyder
       :align: left
       :start: 306
 
-   | If you want to use other modules in Spyder that don't come with our installer, you need to install `Miniconda`_ (**only if you don't have Anaconda or Miniconda yet!**).
+   If you want to use other modules in Spyder that don't come with our installer, you need to install `Miniconda`_ (**only if you don't have Anaconda or Miniconda yet!**).
    For Spyder to recognize it, the installation should be done in one of the following default paths:
 
    .. table::
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -176,7 +176,8 @@ Using Spyder
       :align: left
       :start: 306
 
-   | If you want to use other modules in Spyder that don't come with our installer, you need to install `Miniconda`_ (**only if you don't have Anaconda or Miniconda yet!**). For Spyder to recognize it, the installation should be done in one of the following default paths:
+   | If you want to use other modules in Spyder that don't come with our installer, you need to install `Miniconda`_ (**only if you don't have Anaconda or Miniconda yet!**).
+   For Spyder to recognize it, the installation should be done in one of the following default paths:
 
    .. table::
 
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -176,9 +176,7 @@ Using Spyder
       :align: left
       :start: 306
 
-  |
-
-   If you want to use other modules in Spyder that don't come with our installer, you need to install `Miniconda`_ (**only if you don't have Anaconda or Miniconda yet!**). For Spyder to recognize it, the installation should be done in one of the following default paths:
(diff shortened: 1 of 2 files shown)


juanis2112 (approved):
This seems fixed. Thanks @dalthviz for working on this. None of the failed checks are your fault so I'll approve.

doc/faq.rst
```diff
@@ -176,9 +176,7 @@ Using Spyder
       :align: left
       :start: 306
 
-  |
-
-   If you want to use other modules in Spyder that don't come with our installer, you need to install `Miniconda`_ (**only if you don't have Anaconda or Miniconda yet!**). For Spyder to recognize it, the installation should be done in one of the following default paths:
+   | If you want to use other modules in Spyder that don't come with our installer, you need to install `Miniconda`_ (**only if you don't have Anaconda or Miniconda yet!**). For Spyder to recognize it, the installation should be done in one of the following default paths:
```
CAM-Gerlach, line 179:
```suggestion
   |

   If you want to use other modules in Spyder that don't come with our installer, you need to install `Miniconda`_ (**only if you don't have Anaconda or Miniconda yet!**).
   For Spyder to recognize it, the installation should be done in one of the following default paths:
```

One Line Per Sentance (this does inject some extra vertical space, but that will be fixed by removing the `|` and instead making the CSS change).

CAM-Gerlach (changes requested):
Thanks @dalthviz . Looking at it in more detail, the problem here is twofold. As you're likely aware, PR #277 (specifically bc2e0d5 ) added the block separator at the incorrect indentation level, which was the proximate cause of the bug, and it was not caught when checking the deploy preview. However, I would suggest that instead of the original approach of manually adding in a block separator whenever a video is followed by text (which is a bit of a hack that requires remembering nitpicky syntax like this), we fix it properly in the custom CSS like so:

```
div.video-container-container + p {
	margin-top: 1em;
}
```

Other than the nitpick (which wasn't introduced by this PR, and I can add a lightweight regex check to Pre-Commit to detect and optionally fix in the future), we can either merge this now and I can open another PR to add the above and remove the `|`, or feel free to add that yourself here (perhaps on [line 273](https://github.com/spyder-ide/spyder-docs/blob/a3a579ce6e7e42c9d9bbd88ac4e76650b7093e5c/doc/_static/css/custom_styles.css#L273) of `custom_styles.css` with an appropriate comment).

Thanks!

doc/faq.rst
dalthviz: Hi @CAM-Gerlach glad to see you back! Regarding this suggestion, when I apply this change in my local I see a bigger space that the one you can see with the initial fix of putting the `|` in the correct place (as you describe in your comment):

Initial fix:
image


With the suggestion:

image

However I'm not totally sure but is that actually what you want? (to leave that space) If you put the `|` in the correct place you can have one line per sentence and the correct spacing between the video and the text.

dalthviz: Or maybe I just should do the CSS addition here and remove the `|`?

CAM-Gerlach: Good to see you again too!

> Or maybe I just should do the CSS addition here and remove the `|`?

Sorry for the confusion. Yeah, that was what I intended; this suggestion was an alternative if you wanted to have me do the CSS change in another PR. Making the first sentance a block with the `|` would be okay too, it just adds an extra div around that sentence that adds margin both above and below that doesn't exactly match the following sentence, but either way it doesn't matter since it would be shortly fixed anyway by my followup PR. So up to you. Here's the alternative for reference (not directly tested).

```suggestion
   | If you want to use other modules in Spyder that don't come with our installer, you need to install `Miniconda`_ (**only if you don't have Anaconda or Miniconda yet!**).
   For Spyder to recognize it, the installation should be done in one of the following default paths:
```

CAM-Gerlach (approved):
LGTM now, thanks @dalthviz ! Since we have two :heavy_check_mark: , I verified it works in the deploy previews and this introduces no new check failures, going ahead and merging now.

PR #283, opened 2021-09-18
PR: Don't show plugins in index that aren't yet compatible with Spyder 5

juanis2112:
# Pull Request

## Pull Request Checklist

* [ ] Read and followed this repo's [Contributing Guidelines](https://github.com/spyder-ide/spyder/blob/master/CONTRIBUTING.md)
* [ ] Checked your writing carefully for correct English spelling, grammar, etc
* [ ] Described your changes and the motivation for them below

## Description of Changes

Removed plugins index from docs for spyder 5 until all the plugins are migrated and compatible with Spyder 5.

## Issue(s) Resolved

Fixes #


CAM-Gerlach:
Are some of the plugins compatible yet? If so, it makes sense to still include the plugin index, and instead move the incompatible plugins to a comment in `indexplugins.rst`. Also, I wouldn't suggest waiting until "all" of the plugins are compatible, and rather simply add each plugin as it is updated (and some might never be). No reason to hide documentation for the working plugins, only the non-working ones.

ccordoba12:
> Are some of the plugins compatible yet?

Only Spyder-terminal, but I think @juanis2112's idea is to update all plugin docs at the same time.

> and instead move the incompatible plugins to a comment in indexplugins.rst

That's a good idea. @juanis2112, what do you think?

CAM-Gerlach:
> Only Spyder-terminal, but I think @juanis2112's idea is to update all plugin docs at the same time.

Yeah, but that means that all the plugin docs are blocked on the slowest to migrate plugin, instead of being added whenever the plugin gets migrated. IMO, if and when a plugin is supported, I'd favor making its docs available immediately, with any changes necessary so that the basic instructions are accurate for Spyder 5. If doing the updated screenshots and such is the blocker, that can be deferred, as we did for the Spyder 4 docs, as its better to have working docs with slightly out of date screenshots than no docs at all, assuming the plugin is compatible.

juanis2112:
I think however you think is best is fine for me @CAM-Gerlach. Do you want to do the change?

CAM-Gerlach:
> Do you want to do the change?

Sure.

CAM-Gerlach (approved):
LGTM now, thanks. @juanis2112 All good to merge?

CAM-Gerlach:
Rebase complete; trivial PR, preview looks good and we all agree on how we want this to look, so going ahead and merging.

PR #291, opened 2021-09-29
PR: Clarify line limit and how to clear history log

### LINKED ISSUE #252: Mention the limit of saved lines in the History pane
# Issue Report

## Issue Description

The current limit is 1000 lines but we are mentioning that at the moment. This issue was prompted by the following SO question:

https://stackoverflow.com/questions/67526826/clear-history-tab-of-ipython-console

http://docs.spyder-ide.org/current/panes/historylog.html

We should mention the limit on that section and that it's not possible to change it.

### Other information

    Let's make this change for both Spyder 4 and 5.

    No. @juanis2112, please take care of this one.

CAM-Gerlach:
# Pull Request

## Pull Request Checklist

## Description of Changes

Adds clarifications for the number of lines stored in the history pane, and how to manually clear them.

## Issue(s) Resolved

Fixes #252

### DIFF
--- a/doc/panes/historylog.rst
+++ b/doc/panes/historylog.rst
@@ -2,7 +2,7 @@
 History
 #######
 
-With the **History** pane, you can view all the commands you've entered into any :doc:`ipythonconsole`, along with their timestamp.
+With the **History** pane, you can view the recent code and commands you've entered into any :doc:`ipythonconsole`, along with their timestamp.
 
 .. image:: /images/history/history-standard.png
    :alt: Spyder History Log, displaying a list of previously executed commands
@@ -17,6 +17,7 @@ Navigating the :guilabel:`History` pane is very straightforward.
 Each Spyder session is marked by a date and timestamp, making it easy to remember when you executed a certain command.
 Statements can be selected and copied from the context menu or with the normal system shortcuts.
 Just like in the Editor, selecting a word or phrase displays all other occurrences, and full syntax highlighting is also supported.
+The last ≈1000 lines entered are stored in the pane.
 
 
 
@@ -40,6 +41,8 @@ You might need to show invisible files in order to see it on a non-Windows opera
 
 .. image:: /images/history/history-log-file.png
    :alt: Spyder History Log file
+
+While there is currently no built-in way to clear history from the Spyder interface aside from resetting preferences, you can do so by closing Spyder, deleting this file and restarting Spyder again.
 
 
 


doc/panes/historylog.rst
```diff
@@ -17,6 +17,7 @@ Navigating the :guilabel:`History` pane is very straightforward.
 Each Spyder session is marked by a date and timestamp, making it easy to remember when you executed a certain command.
 Statements can be selected and copied from the context menu or with the normal system shortcuts.
 Just like in the Editor, selecting a word or phrase displays all other occurrences, and full syntax highlighting is also supported.
+The last ≈1000 lines entered are stored in the pane, which is currently not customizable (but may be increased in the future).
```
juanis2112, line 20: I don't see the purpose of mentioning that it may be increased in the future, or that it is not customizable. It makes it feel like it is "wrong" somehow. What do you think?

doc/panes/historylog.rst
```diff
@@ -41,6 +42,8 @@ You might need to show invisible files in order to see it on a non-Windows opera
 .. image:: /images/history/history-log-file.png
    :alt: Spyder History Log file
 
+While there is currently no built-in way to clear history from the Spyder interface aside from resetting preferences, you can do so by closing Spyder, deleting this file and restarting it again.
```
juanis2112, line 45:
```suggestion
While there is currently no built-in way to clear history from the Spyder interface aside from resetting preferences, you can do so by closing Spyder, deleting this file and restarting Spyder again.
```
I know it sounds repetitive but the last noun before this was "this file" so it is a bit confusing. What do you think?

juanis2112 (commented):
Thanks for solving this @CAM-Gerlach! I made two small suggestions.

doc/panes/historylog.rst
CAM-Gerlach: I agree, I would favor not including both of those, but I added the note that it is "not customizable" due to @ccordoba12 's comments on the original issue this is addressing in response to a user asking about clearing it,

> We should mention the limit on that section and that it's not possible to change it.

I then included the "may be increased in the future" as a result of that to help mitigate negative user sentiment surrounding an arbitrarily low limitation that was explicitly stated to be hard-coded, and reflect that it might be modified in the future (as discussed on #252 ).

Ideally we could get Carlos' opinion as the original issue author, but given how busy he is and how relatively minor this is, probably best for us to just decide now. So if you still agree it should be changed, we can go with that:

```suggestion
The last ≈1000 lines entered are stored in the pane.
```

doc/panes/historylog.rst
CAM-Gerlach: Great catch and excellent point, thanks.

doc/panes/historylog.rst
ccordoba12: I agree that for now we shouldn't mention this.

CAM-Gerlach:
Not sure why Netlify isn't firing, but aside from a cursory check of that, this should be ready for approval now @juanis2112

juanis2112 (approved):
Thanks!

CAM-Gerlach:
Thanks for the review, @juanis2112 . Merging...

PR #294, opened 2021-09-30
PR: Optimize loading-critical tour images for Spyder 4

CAM-Gerlach:
# Pull Request

## Pull Request Checklist

## Description of Changes

While unfortunately I (somehow, since I was intimately involved with the PR) didn't catch this initially and thus the images will always take up additional space in the repo, load time is critical for the tour images since it results in significant perceptible delay for the user, blocks the various scripts from firing and in some scenarios, can even cause unpredictable bugs. Running optipng on the images reduced their size by a consistent 30-40% for free, with zero other impacts, so it makes sense to do this here (for the many other unoptimized images, repo space is more critical than load time, and if they were merged unoptimized, we will never be able to recover it and in fact optimizing will only increase the repo size, so I haven't touched them).


CAM-Gerlach:
Since this PR is purely technical and makes no changes to anything user-visible, and there is already a large backlog of PRs waiting on the other maintainers to review (which will only get longer due to me being banned from reviewing non-core-dev PRs), I'll go ahead and merge this once the checks pass and I double-check the preview looks identical to before.

PR #299, opened 2021-10-05
PR: Move sub-indicies to subdirectories where they belong

CAM-Gerlach:
# Pull Request

## Pull Request Checklist

## Description of Changes

When I migrated the docs from a flat (everything in one big directory) to a more organized hierarchical layout, I neglected to also migrate the indecies to where they are supposed to be, as `index.rst`es in their subdirs as Sphinx expects them to be and is convention everywhere else.

In addition to helping keep things more organized following universal convention. match the layout of the built website and more familiar and expected to newer contributors, this means navigating to the subdirectory roots will work correctly, going to the index page instead of erroring out, as users expect (and which will allow us to include useful content on them if we want to in the future, like we do for the overall index page), plus the toctree listings can be simpler and it avoids any edge-case bugs in Sphinx or the theme due to this non-standard layout.

This does change the layout of the sub-index pages, but as they are currently only used internally and not publicly accessible other than by knowing their links, this won't break any inbound links while ensuring future ones will remain stable. The only PR this affects is #283 , which I can trivially rebase once this is merged.


CAM-Gerlach:
Since this is a purely technical change and I've verified in works in the deploy preview, going ahead with this.

PR #339, opened 2022-11-17
Update notebook.rst

Aryo-Zare:
installing the notebook with a different command for spyder 5.

# Pull Request

## Pull Request Checklist

* [ ] Read and followed this repo's [Contributing Guidelines](https://github.com/spyder-ide/spyder-docs/blob/master/CONTRIBUTING.md)
* [ ] Checked your writing carefully for correct English spelling, grammar, etc
* [ ] Described your changes and the motivation for them below

### Issue(s) Resolved

Fixes #

### DIFF
--- a/doc/plugins/notebook.rst
+++ b/doc/plugins/notebook.rst
@@ -19,6 +19,13 @@ If you installed Spyder using conda, the best way to install Spyder-notebook is 
 
    conda install spyder-notebook -c spyder-ide
 
+
+for spyder 5 this command doesn't work. Use conda forge instead :
+
+.. code-block:: bash
+
+    conda install -c conda-forge spyder-notebook
+
 .. important::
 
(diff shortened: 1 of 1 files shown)


CAM-Gerlach (changes requested):
Thanks for reporting this issue.

Aside from needing the follow the above contributiong checklist, the various basic grammar, capitalization, formatting, etc. issues, and the fact that, if correct, this should just replace the existing usage instead since this PR is targeting the Spyder 5 docs, there seems to be a more fundamental concern here. The real issue seems to be that the latest version of Spyder-Notebook wasn't released in the `spyder-ide` channel (which is at 0.3.2), only the main `conda-forge` channel (at 0.4.0). Otherwise, if people aren't using Conda-Forge in their environments already, the resulting mixed channels will likely cause all sorts of problems down the road, and in some cases right away.

@ccordoba12 Can it be updated to 0.4.0+ in the `spyder-ide` channel? If so, this can just be closed.

PR #367, opened 2024-04-25
Adopt PyData-Sphinx-Theme for Spyder Docs

CAM-Gerlach:
# Pull Request

## Pull Request Checklist

* [x] Read and followed this repo's [Contributing Guidelines](https://github.com/spyder-ide/spyder-docs/blob/master/CONTRIBUTING.md)
* [x] Based your PR on the latest version of the correct branch (master or 4.x)
* [x] Checked your writing carefully for correct English spelling, grammar, etc
* [x] Described your changes and the motivation for them below
* [x] Noted what issue(s) this pull request resolves, creating one if needed

## Description of Changes

As has long been planned, this (currently) WiP PR migrates the Spyder Docs site from our very outdated, heavily customized PyData Sphinx Theme fork to the modern upstream versions, as well as migrating our other obsolete dependencies blocked on the same (Sphinx 3, Sphinx-Panels and Sphinx-Multiversion to Sphinx 7, Sphinx-Design and PST's version switcher, respectively).

Currently quite WiP, but publishing this as a PR for visibility and so we have a public preview. The plan is to work on this over the next two weeks, and have it ready to merge before I leave for PyCon.

Migration elements:
* From a very outdated heavily customized fork of the PyData Sphinx Theme to the latest upstream version (0.15.2)
* From the ancient Sphinx 3 to the latest Sphinx 7
* From the obsolete, unsupported Sphinx-Panels to its successor Sphinx-Design
* From the unmaintained, no longer properly functional Sphinx-Multiversion to the built-in multiversion support of PST

As a result, this:
* Substantially improves the site's layout, responsiveness and functionality,
* Allows us to take advantage of current and future improvements in the theme,
* Adds critical support for multi-versions and translations that we need,
* Eliminates a large amount of complexity and maintenance on our site,
* Gets rid of obsolete dependencies in favor of modern, supported equivalents
* Overall, pays off a large amount of accumulated techdebt

## Work plan for initial PR

* [x] Commit initial baseline
    * [x] Update requirements* files
    * [x] Update contributing guide
* [x] Fix build things
    * [x] Image permissions issue
    * [x] Language tag
    * [x] Theme settings warning
    * [x] Panel directive on main page
    * [x] Dropdown ids
* [x] Fix CSS and JS
    * [x] Div to section: CSS placement and color of fa icons
        * [x] Check others (videos, etc)
    * [x] Sphinx-Panels to Sphinx-Design CSS classes/structure (panels, dropdown)
        * [x] Dropdown (FAQ)
[...]

PR #374, opened 2024-07-26
Update version warning banner to always point to the latest stable docs

CAM-Gerlach:
# Pull Request

## Pull Request Checklist

## Description of Changes

Currently, the warning banner for old (and new) versions embeds a link title and URL that hardcode the version number of the latest stable docs. With Sphinx-Multiversion that was fine, since all branches were rebuilt in sync so even the old banners would always be up to date, but with the new theme introduced in #367 , this will no longer be the case and unmaintained branches will be static, and not migrated to the new theme. Therefore, we need to update the old banner to be "evergreen" and point to the `/current/` version rather than a hardcoded version value.

I'll also fix any linkcheck issues that may come up.

Required for #367


ccordoba12 (approved):
Thanks @CAM-Gerlach!

PR #377, opened 2024-09-03
Migrate docs infrastructure from Make to Nox

CAM-Gerlach:
# Pull Request

## Pull Request Checklist

## Description of Changes

Our automation infrastructure still uses a bunch of old Makefiles, including the soft-deprecated Sphinx ones and our own in the root, that don't work cross-platform, are arduous to edit and don't provide Python-specific facilities of dedicated tools. Therefore, this PR ports our existing Makefile functionality (and quite a bit more) to Nox, updates the Readme/Contributing guide and CIs accordingly, and removes the old ones.


CAM-Gerlach:
This is ready for review; the one remaining blocker is a just-emerged issue, RJWadley/stylelint-no-unsupported-browser-features#299 in stylelint-no-unsupported-browser-features due to a change in the CanIUse data it consumes dynamically from the web, that causes Stylelint proper to crash.

Seems its likely to be fixed shortly as a PR by the maintainers that fixes it, anandthakker/doiuse#191 , is already open, but it would also block merge of the translations infra PR if it isn't fixed by then, so we can wait until I have that up and then just disable that plugin temporarily if its still an issue by the time its hard-blocking.

ccordoba12 (approved):
Looks good to me, thanks @CAM-Gerlach!

ccordoba12:
> so we can wait until I have that up and then just disable that plugin temporarily if its still an issue by the time its hard-blocking.

I think that's not necessary because `stylelint` runs a js/css lint check and this PR hasn't changed that. So please disable it so this one can be merged later today.

CAM-Gerlach:
> I think that's not necessary because stylelint runs a js/css lint check and this PR hasn't changed that. So please disable it so this one can be merged later today.

That's not why I was suggesting delaying it, but rather to just avoid the messyness of disabling a check in an unrelated PR, and then remembering it to enable it again in another unrelated PR. Also, it turns out that we're going to need to fix the broken links in the docs, which does involve some non-trivial CSS changes to add the Linux icon to the download buttons.

However, after reading the issue again, it might actually be workaroundable by pinning a specific transitive sub-dependency with pre-commit, which is not so bad for now and now that we have to do it in the link PR anyway, that makes more sense since it's all about fixing links and things that rotted recently.

> You can enable it again in your next PR, which will probably change and/or add js/css for the new translations dropdown.

As we've discussed many times by now :) , my next PR will not be adding the translation dropdown, but first adding the translation _infrastructure_ that is a prerequisite, to automate extracting the translations, interfacing with Crowdin and building with them (and actually doing so).

PR #384, opened 2024-09-24
Add Nox setup & setup-remotes commands + other improvements

CAM-Gerlach:
# Pull Request

## Pull Request Checklist

* [x] Read and followed this repo's [Contributing Guidelines](https://github.com/spyder-ide/spyder-docs/blob/master/CONTRIBUTING.md)
* [x] Based your PR on the latest version of the correct branch (master or 4.x)
* [x] Checked your writing carefully for correct English spelling, grammar, etc
* [x] Described your changes and the motivation for them below

## Description of Changes

The new `setup-remotes` action not only sets up the `origin` and `upstream` remotes in both the recommended `fetch` and `push` configuration by connection method (HTTPS or SSH), but also detects and offers to fix issues with the existing config.

- [x] Implement final improvements and fixes to Nox infra
    - [x] Add setup Nox action to set up upstream remote (and fetch), install hooks and clean build dir
    - [x] Add setup action to Contributing Guide
    - [x] Add docs alias for build commands (and also change _docs)
    - [x] Also change `run` to error if no args passed
    - [x] Check for and fix other issues

PR #383, opened 2024-09-22
Update pre-commit hooks and config for 2024-09

CAM-Gerlach:
# Pull Request

## Pull Request Checklist

* [x] Read and followed this repo's [Contributing Guidelines](https://github.com/spyder-ide/spyder-docs/blob/master/CONTRIBUTING.md)
* [x] Based your PR on the latest version of the correct branch (master or 4.x)
* [x] Checked your writing carefully for correct English spelling, grammar, etc
* [x] Described your changes and the motivation for them below

## Description of Changes

Ended up needing to investigate and resolve a bunch of issues due to a number of major version bumps, breaking changes and repo retirements/successions, but that all only needs to be done once, and it should hopefully be smoother sailing in the future.

- [x] Update Pre-Commit hooks to latest to prepare for porting to Spyder-API-Docs
    - [x] Do auto-update and ensure all hooks work and pass
        - [x] Fix issues with auto-update (pre-release versions, pinned versions)
        - [x] Re-investigate caniuse/browser config problem
        - [x] Investigate and fix problems with Prettier
        - [x] Migrate from old unmaintained eslint-config-standard to neostandard
        - [x] Migrate eslint config and plugins to v9 and new eslint JS config file
        - [x] Migrate stylelint and plugins to v16
        - [x] Test and fix other hook issues
    - [x] Check changed hooks and update config where needed
        - [x] And remove no-var config for ESlint
    - [x] Update Pre-Commit action and update .gitignore per spyder-ide/spyder-website#13

PR #385, opened 2024-09-25
Remove Curlylint, add Black and fix ESLint

CAM-Gerlach:
# Pull Request

## Pull Request Checklist

## Description of Changes

Followup from #383 and #384 , removes the unused, unmaintained and slow Jinja template linter Curlylint, adds the code formatter Black and applies the latter to blacken the repo's Python scripts (excluding the examples in the `/workshops` subdirectory). Additionally, adds a custom `.black.toml` config file to avoid adding an unnecessary, inappropriate and potentially misleading `pyproject.toml` to this repo that's not any form of Python package.


CAM-Gerlach:
Seems like there's a strange, hopefully temporary dependency conflict issue trying to install eslint (never seen that before), so hopefully that'll result itself in a day or two; there's no rush to merge this immediately.

CAM-Gerlach:
Welp, seems like the strange NPM dependency conflict that suddenly popped up has still not been resolved, so I'll dig in and investigate it myself.

CAM-Gerlach:
Welp, after several hours investigating, I reported this as eslint/eslint#18956 , and appears to be solved by eslint/eslint#18945 and was able to drill down to a workaround for now. Unfortunately, pinning the specific problematic dep didn't work for some reason, so I had to pin ESLint on a modestly older version for now until that's released.

ccordoba12 (commented):
Looks good to me, but I think one thing is missing before merging: we should add commit 9e75d1834b30a486e6b8d0a011be43b587e8cd2d to `.git-blame-ignore-revs` so that it's ignored by `git blame`.

That's the recommended approach when applying Black to a repo and I think it's a good idea.

CAM-Gerlach:
Okay, sure; I didn't add that before since the blackening only affects a couple scripts, and users need to either manually pass the path to that file every time, or (in recent-ish Git versions) manually configure Git to set and use this as the ignore file. However, after thinking about it a bit, I realized I can just add that to the things the `setup` command does, which I did so.

I also noticed and fixed a small perf issue in `setup-remotes`.

ccordoba12 (approved):
Looks good to me now, thanks @CAM-Gerlach!

